PR #1, opened 2023-02-07
[ORCA-129] Create Python package structure based on `py-dcqc`

BrunoGrandePhD:
I have come to regret some design decisions that I made when I was working on the `sagetasks` proof of concept. When I worked on the DAG for launching a workflow on Cavatica, I found the process awkward in Airflow. We also should 

I could create a backwards-incompatible release (v2), but I've also been thinking about using the term "sage" without "bionetworks", which has proved to be problematic in the past because of trademarks. So, I figured it was perhaps a good time to create the `orca` package. This also gives me the chance to update the Python package structure to leverage all of the small tweaks that I incorporated into `py-dcqc`. 

I'm creating this PR mainly to keep my next PR as minimal as possible. I don't expect folks to review this PR very thoroughly since it leverages what I already have working for `py-dcqc`.


setup.py
```diff
@@ -0,0 +1,21 @@
...
+"""
+from setuptools import setup
+
+if __name__ == "__main__":
+    try:
+        setup(use_scm_version={"version_scheme": "no-guess-dev"})
+    except:  # noqa
+        print(
+            "\n\nAn error occurred while building the project, "
+            "please ensure you have the most updated version of setuptools, "
```
thomasyu888, line 17: Might be know what version of setuptools it should at least be

thomasyu888 (approved):
🚀 LGTM!

setup.py
BrunoGrandePhD: This came with the PyScaffold template, so I don't know what is the minimum version for this package.

PR #2, opened 2023-02-07
[ORCA-128] Design an Airflow-friendly structure for client and tasks code

BrunoGrandePhD:
The goal of this PR was to design the basic structure of a Python module for a service that we want to support in `orca`. I used SevenBridges as the representative service here because it's the focus of the ticket that I'm working on (ORCA-22). 

I ended up implementing the classes because I was happy with the structure. I also maxed out the test and docstring coverage to 100%. I also decided to start documenting design decisions in the [CHANGELOG](https://github.com/Sage-Bionetworks-Workflows/py-orca/pull/2/files#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4ed) so they can be more easily found later. 

**Edit:** For the tests to pass, I'll need to create a Cavatica service account and create a repository secret with the connection URI. 

- [x] Create Cavatica service account
- [x] Store Cavatica account credentials in LastPass
- [x] Create a repository secret with the Cavatica connection URI

Depends on #1 

## Example DAG

Out of curiosity, I've rewritten the [`cavatica-launch-poc.py`](https://github.com/Sage-Bionetworks-Workflows/orca-recipes-airflow/blob/393c904e5a21967bcecbb84447c5bc03de76d856/dags/cavatica-launch-poc.py) DAG that I wrote as if I had already implemented the relevant methods on the `SevenBridgesTasks` class. This is a fictitious example, so I don't know if it'll work as is. For example, the ability to pull in the DAG parameters by including `params` in the task function arguments is based on this [short section](https://airflow.apache.org/docs/apache-airflow/stable/core-concepts/taskflow.html#context) in the Airflow docs, so I'll need to make sure this works. Otherwise, we'll just need to rely on Jinja templating instead (_e.g._ `hook = SevenBridgesHook("{{ params.conn_id }}")`). 

```python
from airflow.decorators import dag, task
from airflow.models.param import Param
from airflow.sensors.base import PokeReturnValue
from orca.services.sevenbridges import SevenBridgesHook

# The `Param` class allows you to define default values and perform basic type validation
params = {
    "conn_id": Param("cavatica_include", type="string"),
    "app_id": Param("bgrande/include-sandbox/kfdrc-rnaseq-workflow-3", type="string"),
}

dag_args = {
    "params": params,
    # ...
}


@dag(**dag_args)
def cavatica_launch_poc():
    @task
    def prepare_task(params=None):
        # Cavatica project ID is already specified by the connection
        hook = SevenBridgesHook(params["conn_id"])
        app_id = params["app_id"]
        task_inputs = {...}
        task = hook.tasks.get_or_create_task(app_id, task_inputs, "cavatica_launch_poc")
        return task.name

    @task
    def launch_task(task_name, params=None):
        hook = SevenBridgesHook(params["conn_id"])
        task = hook.tasks.get_task(task_name)
        task.run()
        return task_name

    # @task.sensor was added in Airflow 2.5
    @task.sensor(poke_interval=300, timeout=604800, mode="reschedule")
    def monitor_task(task_name, params=None):
        hook = SevenBridgesHook(params["conn_id"])
        task_state = hook.tasks.get_task_state(task_name)
        is_done = task_state in {"COMPLETED", "FAILED"}
        # PokeReturnValue allows you to return a value in addition to unblocking the sensor
        return PokeReturnValue(is_done, task_state)

    draft_task_name = prepare_task()
    task_name = launch_task(draft_task_name)
    task_state = monitor_task(task_name)
```


.env.example
```diff
@@ -0,0 +1,65 @@
...
+
+# Synapse credentials
+# -------------------
+# <access-token>: Replace with a Synapse personal access
+#   token (PAT), which can be generated here:
+#   https://www.synapse.org/#!PersonalAccessTokens
+#
+# Examples:
+#   SYNAPSE_CONNECTION_URI=syn://:eyJ0[...]QP7g@
+SYNAPSE_CONNECTION_URI=syn://:<access-token>@
```
thomasyu888, line 29: Hmm, did we want to support the different synapse endpoints?  (That said, I think this is fine - there are 4 separate synapse endpoints which would complicate things)

CHANGELOG.md
```diff
@@ -1,7 +1,20 @@
...
+## Unreleased
 
-- Feature A added
-- FIX: nasty bug #1729 fixed
-- add your changes here!
+### New Features
+
+- Add basic support for SevenBridges service (_i.e_ client factory, tasks, and hook)
+
+### Design Decisions
```
thomasyu888, line 9: I like this, we should try to leverage github releases as much as possible with these things. You can auto generate release notes which will capture all the PR's from the new release to the latest release

src/orca/services/sevenbridges/client_factory.py
```diff
@@ -0,0 +1,174 @@
...
+        """Resolve SevenBridges credentials based on priority.
+
+        This method will update the attribute values (if applicable).
+        """
+        # Short-circuit method if absent because Connection is slow-ish
+        if os.environ.get(self.CONNECTION_ENV) is None:
+            return
+
+        # Retrieve information from environment (if available)
+        env_connection_uri = os.environ.get(self.CONNECTION_ENV, "")
```
BWMac: Is there a reason to include the case for `""` if `os.environ.get(self.CONNECTION_ENV)` is `None` if you have already handled that case above?

src/orca/services/sevenbridges/client_factory.py
```diff
@@ -0,0 +1,174 @@
...
+from typing import Any, Optional
+
+from airflow.models.connection import Connection
+from sevenbridges import Api
+from sevenbridges.http.error_handlers import maintenance_sleeper, rate_limit_sleeper
+
+from orca.errors import ClientArgsError, ClientRequestError
+
+
+@dataclass
```
BWMac: Is the primary advantage to using a `dataclass` here being able to be picky about typing of arguments used to initialize it?

src/orca/services/sevenbridges/hook.py
```diff
@@ -0,0 +1,48 @@
...
+        """Construct SevenBridgesHook using an Airflow connection.
+
+        Args:
+            conn_id: An Airflow connection ID.
+                Defaults to ``default_conn_name``.
+        """
+        super().__init__(*args, **kwargs)
+        self.connection = self.get_connection(conn_id)
+        extras = self.connection.extra_dejson
+        self.project = extras.get("project")
```
BWMac, line 32: is "project" always a required field in the `extras` where project-specific info will live? Might want to document somewhere what should/shouldn't go in there.

.github/workflows/ci.yml
```diff
@@ -63,8 +63,6 @@ jobs:
         platform:
         - ubuntu-latest
         - macos-latest
-        # TODO: Debug the Windows issues
-        # - windows-latest
```
thomasyu888, line 67: Is it worth locking down specific versions of ubuntu/mac/windows?

setup.cfg
```diff
@@ -49,6 +49,7 @@ python_requires = >=3.8
 # Updates here should be reflected in `docs/requirements.txt`
 install_requires =
     apache-airflow~=2.4
+    sqlalchemy<2.0
```
BWMac: Is this a necessary requirement?

BWMac (commented):
I think this is a great step towards having an airflow-friendly package. I especially like the structure of the `SevenBridgesClientFactory` and I think this will be a reuseable model for adding more services

src/orca/services/sevenbridges/client_factory.py
BrunoGrandePhD: The main advantage of the `dataclass` in the standard Python library is that it eliminates boilerplate (like defining an `__init__()` method that just stores the parameters as attributes. They intentionally avoid using the type hints for validation ([source](https://docs.python.org/3/library/dataclasses.html#class-variables)). 

That said, your comment prompted me to look into how easy it would be to leverage `pydantic` validation using type hints, and I found that they have a `dataclass` [drop-in replacement](https://docs.pydantic.dev/usage/dataclasses/). I've made the switch, so we're now validating the input values.

setup.cfg
BrunoGrandePhD: I added a comment to clarify why I added this. It's in response to a warning from SQLAlchemy ([source](https://sqlalche.me/e/b8d9)).

src/orca/services/sevenbridges/client_factory.py
BrunoGrandePhD: You're right! The above short circuit was a relatively late change, so I don't need to use `get()` anymore.

src/orca/services/sevenbridges/hook.py
BrunoGrandePhD: In this case, I'm using the `project` name since that's what is used by SevenBridges. We would use `workspace` for Tower. I added the following to the description of the `hook.py` submodule. Let me know if this answers your questions. 

> This hook also takes advantage of
> extra field associated with connection objects to configure the
> underlying client and/or tasks objects. Whether to include a value
> as the extra connection field probably depends on whether that
> value is useful to change the overall state of the underlying
> client and/or tasks objects (e.g., `project` for SevenBridgesTasks)
> and whether the value is invariable for a given DAG.

src/orca/services/sevenbridges/hook.py
```diff
@@ -0,0 +1,48 @@
...
+
+    @cached_property
+    def client(self) -> Api:
+        """Retrieve authenticated SevenBridges client."""
+        client_args = SevenBridgesClientFactory.map_connection(self.connection)
+        factory = SevenBridgesClientFactory(**client_args)
+        return factory.get_client()
+
+    @cached_property
+    def tasks(self) -> SevenBridgesTasks:
```
thomasyu888, line 46: I think this makes sense.  Just to make sure I'm on the same page, the hook uses the `SevenBridgesClientFactory` to create the client, then the `SevenBridgesTasks` itself takes in the client and has all the commands that the sbg client can do on a specific project.

Is there a way to create projects?

thomasyu888 (approved):
🚀 See my initial comments, but this looks good.   Do you think we would need a `BaseFactory` class?

.env.example
BrunoGrandePhD: That idea did come up in conversation with @BWMac on Monday. This way, we wouldn't be artificially incrementing Synapse IDs (assuming separate databases), and the files would get deleted automatically (I think). This wouldn't work though for tests that use Synapse files as input unless we create them as part of the setup for the Synapse integration tests. 

To address the fact that Synapse actually has 4 different API endpoints, we can use `connection.host = "staging.synapse.org"` as a shorthand to run [this method](https://github.com/Sage-Bionetworks/synapsePythonClient/blob/c0acef913f51ec99c57031077a9273cc76cbb820/synapseclient/client.py#L302) on the Synapse client.

CHANGELOG.md
BrunoGrandePhD: Yeah, minting releases going forward will just involve creating a tagged commit that changes the CHANGELOG heading from "Unreleased" to a specific version and then copy-pasting the contents into the GitHub release notes.

src/orca/services/sevenbridges/hook.py
BWMac: I think this is good, thanks!

.github/workflows/ci.yml
BrunoGrandePhD: I don't see a strong motivation because I would like to know if the CI breaks when the OS is upgraded. On the other hands, we could test a range of versions (probably with Ubuntu), but I'm not very concerned about this.

src/orca/services/sevenbridges/hook.py
BWMac: Looks like there is a way to create projects in the `sevenbridges` [python package](https://sevenbridges-python.readthedocs.io/en/latest/sevenbridges.models/?highlight=project#sevenbridges.models.project.Project.create)

BrunoGrandePhD: The beautiful thing about this setup (_i.e._ giving the hook both the `client` and `tasks` properties) is that if the `tasks` object doesn't offer a convenience function for a given operation, you can always use the `client` object to make "lower-level" calls. 

For example, `SevenBridgesTasks` doesn't have a convenience method for creating a project, but the `sevenbridges.Api` class does ([docs](https://sevenbridges-python.readthedocs.io/en/latest/quickstart/#project-methods-an-introduction-to-methods-in-the-sevenbridges-python-library)). In an Airflow DAG, it would still be one line of code because the SevenBridges client offers an easy method for creating a project. 

```python
@task
def create_project(task_name, params=None):
    hook = SevenBridgesHook(params["conn_id"])
    project = hook.client.projects.create(name="My new project", billing_group=params["billing_group"])
    return project.id
```

thomasyu888: Excellent!

CHANGELOG.md
thomasyu888: Hm, I think we are saying the same thing, but just to be sure.  I'm suggesting we use the github auto feature that creates something like this:.  That said, this feature doesn't keep track of design decisions so I find a lot of value in the CHANGELOG.
 
image

src/orca/services/sevenbridges/client_factory.py
```diff
@@ -0,0 +1,176 @@
+import os
+from dataclasses import field
+from functools import cached_property
+from typing import Any, Optional
+
+from airflow.models.connection import Connection
+from pydantic.dataclasses import dataclass
```
thomasyu888, line 7: Nice usage of pydantic which seems to be regularly maintained.  I know i've said this before, but the more dependencies we add, the more risk we create for ourselves when handling updates.  That said, the dependencies you add always provide large value.

.env.example
thomasyu888: Good idea - I dont think it's urgent to add it but it might be nice for consistency.

The other issue is that the staging endpoint actually isn't always up and running.

src/orca/services/sevenbridges/hook.py
BrunoGrandePhD: I posted my comment before seeing yours, @BWMac. Good catch!

src/orca/services/sevenbridges/client_factory.py
BrunoGrandePhD: Noted! I do try to be judicious when adding dependencies, but I still want to hear if ever you conclude that value added is too limited to justify the dependency.

CHANGELOG.md
BrunoGrandePhD: I'll look into how I can combine the GitHub feature, which is neat, with the CHANGELOG. We should also strive for informative PR titles.

BrunoGrandePhD:
@thomasyu888: Regarding the creation of a `BaseClientFactory` class, you've read my mind. I'm going to implement that when we tackle the second client factory because there is indeed a good amount of shared logic between services for client construction.

PR #3, opened 2023-02-08
[ORCA-130] Implement basic `SevenBridgesTasks` methods for task monitoring

BrunoGrandePhD:
The goal of this PR is to implement the methods necessary to implement an Airflow DAG that can launch a Cavatica task/workflow and monitor it while it's running. It's relatively self-explanatory and borrows a lot from `sagetasks`. Note that I was able to maintain test and docstring coverage at 100%. 

Depends on #2 

## (Updated) Example DAG

I've updated the DAG that I included in #2 based on what I implemented in this PR. It's very similar, but I was able to remove the first task because I have a high-level `create_task()` method which drafts and launches a task in one go. I also simplified the `monitor_task()` Airflow task. 

```python
from airflow.decorators import dag, task
from airflow.models.param import Param
from airflow.sensors.base import PokeReturnValue
from orca.services.sevenbridges import SevenBridgesHook

params = {
    "conn_id": Param("cavatica_include", type="string"),
    "app_id": Param("bgrande/include-sandbox/kfdrc-rnaseq-workflow-3", type="string"),
}

dag_args = {
    "params": params,
    # ...
}


@dag(**dag_args)
def cavatica_launch_poc():
    @task
    def create_task(params=None):
        hook = SevenBridgesHook(params["conn_id"])
        task_inputs = {...}
        task_id = hook.tasks.create_task("{{ run_id }}", params["app_id"], task_inputs)
        return task_id

    @task.sensor(poke_interval=300, timeout=604800, mode="reschedule")
    def monitor_task(task_name, params=None):
        hook = SevenBridgesHook(params["conn_id"])
        task_status, is_done = hook.tasks.get_task_status(task_name)
        return PokeReturnValue(is_done, task_status)

    task_id = launch_task()
    task_final_status = monitor_task(task_id)
```


src/orca/services/sevenbridges/tasks.py
```diff
@@ -1,17 +1,48 @@
...
+        if self.project is None:
+            name = method.__name__
+            message = f"`project` attribute must be set before calling `{name}()`."
+            raise OptionalAttrRequiredError(message)
+        return method(self, *args, **kwargs)
+
+    return wrapper
+
+
+# TODO: Should I change the class suffix to "Ops" to avoid confusion
```
BWMac, line 36: I had a moment of confusion about that when reviewing the other PR, so I would be in favor of this change.

BWMac, line 37: I had a moment of confusion about that when reviewing the other PR, so I would be in favor of this change. I think it gets even more confusing with the addition of new methods below that have the work "task" in their names (even with the addition of your note in the `SevenBridgesTasks` docstring).

src/orca/services/sevenbridges/tasks.py
```diff
@@ -1,17 +1,48 @@
...
+#       with the concept of tasks in SevenBridges?
 @dataclass
 class SevenBridgesTasks:
```
BWMac, line 40: I like the minimized return values in the new methods, definitely makes it more airflow-friendly

BWMac (commented):
Looks great!

src/orca/services/sevenbridges/tasks.py
BrunoGrandePhD: I'll make this change in a separate PR

src/orca/services/sevenbridges/tasks.py
thomasyu888: +1

src/orca/errors.py
```diff
@@ -11,3 +11,11 @@ class ClientRequestError(OrcaError):
 
 class ClientArgsError(OrcaError):
     """Client arguments are missing or invalid."""
+
+
+class OptionalAttrRequiredError(OrcaError):
```
thomasyu888, line 16: Nice usage of custom errors here. Just for @BWMac , we always want to use the python built in exceptions as much as possible first prior to building our own (like we did here, since there no built in exceptions that describes these custom errors).

src/orca/services/sevenbridges/tasks.py
```diff
@@ -46,3 +77,103 @@ def from_creds(
...
+    def launch_task(self, task_id: str) -> str:
+        """Launch a draft task (workflow run).
+
+        Args:
+            task_id: Task ID.
+
+        Returns:
+            The input task ID (passed through).
+        """
+        task = self.client.tasks.get(task_id)
```
thomasyu888, line 150: Does this just cause an error when a draft task doesn't exist?

thomasyu888 (approved):
🌻 this is awesome. Thanks for creating convenience functions that makes it easier for usage with airflow dags.

src/orca/services/sevenbridges/tasks.py
BrunoGrandePhD: I just checked, and this is indeed the case. 

```python
>>> client.tasks.get("foo")
NotFound: Requested task does not exist.
```

PR #4, opened 2023-02-09
[ORCA-131] Change "Tasks" suffix to "Ops" to avoid confusion

BrunoGrandePhD:
In the previous PRs, the term "task" became confusing because it had multiple meanings. In light of this, I opted to change the "Tasks" suffix to "Ops" (short for operations). This PR introduces no functional change. 

1. In the `SevenBridgesTasks` class, it referred to the class methods
2. In Airflow, it refers to steps in a DAG (especially using the TaskFlow API)
3. In SevenBridges, it refers to workflow runs


thomasyu888 (approved):
✏️ Nice update - I agree that this will change will reduce confusion in the future.

PR #5, opened 2023-02-09
[ORCA-133] Configure `orca` as an Airflow provider package

BrunoGrandePhD:
This allows the Airflow hooks to be discoverable by Airflow. One benefit of this is that it allows for the "SevenBridges" connection type in the UI, which can be useful for testing. 

image


thomasyu888 (approved):
⚡ Looks good!  I'm looking up more about extra-links - looks very intersting!

PR #6, opened 2023-02-10
[ORCA-134] Update "orca" based on Airflow testing

BrunoGrandePhD:
I was performing some testing in Airflow, and I wasn't able to get things to run from start to finish. So, I decided to write an acceptance test that emulates the DAG but without Airflow. 

I had to change a few things about how the classes interact with each other in order to allow the hook to be instantiated outside of an Airflow context, but I got it to work with minimal changes! This is great from a testing perspective! 

You'll notice that I also added `vulture` as a pre-commit hook and development dependency. I wanted to make sure that my changes didn't result in dead (_i.e.) unused) code.

### DIFF
--- a/src/orca/services/sevenbridges/client_factory.py
+++ b/src/orca/services/sevenbridges/client_factory.py
@@ -1,14 +1,18 @@
+from __future__ import annotations
+
 import os
 from dataclasses import field
 from functools import cached_property
-from typing import Any, ClassVar, Optional, Type
+from typing import TYPE_CHECKING, Any, ClassVar, Optional, Type
 
-from airflow.models.connection import Connection
 from pydantic.dataclasses import dataclass
 from sevenbridges import Api
 from sevenbridges.http.error_handlers import maintenance_sleeper, rate_limit_sleeper
 
 from orca.errors import ClientArgsError, ClientRequestError
+
+if TYPE_CHECKING:
+    from airflow.models.connection import Connection
 
 
 @dataclass(kw_only=False)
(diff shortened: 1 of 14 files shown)


src/orca/services/sevenbridges/client_factory.py
```diff
@@ -80,9 +84,34 @@ def map_connection(connection: Connection) -> dict[str, Any]:
...
         return kwargs
 
+    @classmethod
+    def connection_from_env(cls) -> Connection:
+        """Generate Airflow connection from environment variable.
+
+        Returns:
+            An Airflow connection
+        """
+        # Following Airflow's lead on this non-standard practice
```
BWMac, line 98: I saw [this](https://airflow.apache.org/docs/apache-airflow/stable/best-practices.html) the other day and was starting to think about using this best practice. Nice!

src/orca/services/sevenbridges/client_factory.py
```diff
@@ -80,9 +84,34 @@ def map_connection(connection: Connection) -> dict[str, Any]:
...
+        # because this import does introduce a bit of overhead
+        from airflow.models.connection import Connection
```
BWMac, line 100: I saw [this](https://airflow.apache.org/docs/apache-airflow/stable/best-practices.html) the other day and was starting to think about using this best practice. Nice!

BWMac (approved):
LGTM! This is awesome and another step towards having plug-and-play functionality in airflow

PR #7, opened 2023-02-10
Fix inconsistent package name

BrunoGrandePhD:


BrunoGrandePhD:
I confirmed that this fixes the issue that I observed. Merging without review since this is a hotfix.

PR #8, opened 2023-02-28
[ORCA-151] Create base class for client factories

BrunoGrandePhD:
Before we add Nextflow Tower as a service to Orca, I wanted to extract the common functionality from the SevenBridges client factory into a base class that can be shared. In doing so, I identified the opportunity for a "service config" class for a better separation of concerns. I also created a base class for the service config. 

You'll notice that I've upgrade mypy to 1.0. They released v1 earlier in February, and it includes support for a new typing construct called `Self`. You'll see it used in this PR. It allows me to create constructor class methods with type hints. While `Self` was added to Python 3.11, the `typing_extensions` package backports this functionality to older Python versions. 

I'll be opening up separate PRs for the base classes for ops and hooks. 

**Edit:** Just as a heads-up, I opened #11, which has several improvements to the `BaseConfig` and how the other base classes interact with it. Unfortunately, that refactor started on top of #10, so I'm unable to easily merge those changes into this PR. Please review this as usual, and I'll resolve any comments that I think are addressed in #11.


src/orca/errors.py
```diff
@@ -9,8 +9,8 @@ class ClientRequestError(OrcaError):
     """Client request failed."""
 
 
-class ClientArgsError(OrcaError):
-    """Client arguments are missing or invalid."""
+class ClientAttrError(OrcaError):
```
thomasyu888, line 12: Nit: I don't mind this custom error, but we should try to leverage python base error as much as possible.  This seems like it could use `AttributeError`.

src/orca/services/base/config.py
```diff
@@ -0,0 +1,71 @@
...
+
+@dataclass(kw_only=False)
+class BaseServiceConfig(ABC):
+    """Simple container class for service-related configuration."""
+
+    connection_env_var: ClassVar[str]
+
+    @classmethod
+    @abstractmethod
+    def from_connection(cls, connection: Connection) -> Self:
```
thomasyu888, line 22: Is `TYPE_CHECKING` ever false?  If so, would there be an error here where `Connection` isn't found.

src/orca/services/base/config.py
```diff
@@ -0,0 +1,71 @@
...
+        """
+        # Following Airflow's lead on this non-standard practice
+        # because this import does introduce a bit of overhead
+        from airflow.models.connection import Connection
+
+        env_connection_uri = os.environ.get(cls.connection_env_var)
+        return Connection(uri=env_connection_uri)
+
+    @classmethod
+    def is_env_available(cls) -> bool:
```
thomasyu888, line 65: Why is this a class method?

thomasyu888 (approved):
🛸 This makes sense and is clean.  I did some testing:

```
from orca.services.sevenbridges.client_factory import (
    SevenBridgesClientFactory,
    SevenBridgesConfig,
)
from orca.services.sevenbridges.ops import (
    SevenBridgesOps,
)
auth_token = ""
config = SevenBridgesConfig(
    api_endpoint="https://cavatica-api.sbgenomics.com/v2",
    auth_token=auth_token,
    project="orca-service/test-project"
)

factory = SevenBridgesClientFactory.from_config(config)
client = factory.get_client()
ops = SevenBridgesOps(client, config.project)

ops.client.tasks.query(project="orca-service/test-project")
ops.get_task(name="bgrande - 2023-02-09 14.38.44 - b5c482aa-06b1-4ba0-97eb-67135081962f",
             app_id="orca-service/test-project/kfdrc-rnaseq-workflow/0")
```

Just some minor comments and questions, but going to pre-approve.  I'd give @BWMac some time to look through these for learning.

src/orca/errors.py
BrunoGrandePhD: I forgot about `AttributeError`. I agree that this would be a better exception to use (or inherit from for a more specific custom exception class). That said, I updated this class in #11 to `ConfigError` ([source](https://github.com/Sage-Bionetworks-Workflows/py-orca/pull/11/files#diff-a96ad2c0eb692c5a8d2370cb1dfaca8d2afea0f7b72a4873dd054dcf632fce38L12-R12)). I'll update that exception class to inherit from `AttributeError`.

src/orca/services/base/config.py
BrunoGrandePhD: It's false when the code is being run by a user. It's true when the code is being scanned by mypy or Pylance (VS Code). This isn't an error thanks to the `from __future__ import annotations` import statement, which causes type annotations to be parsed as strings ([PEP 563](https://peps.python.org/pep-0563/)).

src/orca/services/base/config.py
BrunoGrandePhD: Good question. I opted to make this a class method because (1) it doesn't interact (read or write) with any instance attributes and (2) I want it to be callable from class object without requiring an instance to exist.

BWMac (approved):

PR #9, opened 2023-02-28
[ORCA-152] Create base class for ops

BrunoGrandePhD:
Most of the functionality in the `SevenBridgesOps` class is specific to SevenBridges, so the base Ops class is small. 

I took the opportunity to remove support for an SevenBridgesOps instance without a `project`. This is an unlikely use case, so I'm dropping support.

### DIFF
--- a/src/orca/services/sevenbridges/ops.py
+++ b/src/orca/services/sevenbridges/ops.py
@@ -1,44 +1,21 @@
 from __future__ import annotations
 
-from functools import wraps
 from typing import Any, Optional, cast
 
 from pydantic import ConfigDict
 from pydantic.dataclasses import dataclass
 from sevenbridges import Api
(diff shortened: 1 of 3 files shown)


src/orca/services/sevenbridges/ops.py
```diff
@@ -47,23 +24,25 @@ class SevenBridgesOps:
     """
 
     client: Api
-    project: Optional[str] = None
+    project: str
```
thomasyu888, line 27: This makes sense. Most interactions with SevenBridgesOps is going to be around a project (as it pertains to the Hook).  As you've mentioned previously, the beauty of this implementation is that you can still interact with all of the seven bridges API.

```
ops.client.projects.query()
```

thomasyu888 (approved):
🚀 LGTM!

BWMac (approved):

PR #10, opened 2023-02-28
[ORCA-153] Create base class for hooks

BrunoGrandePhD:
The hook class was an excellent candidate for creating a base class. Most of its functionality is service-agnostic beyond its reliance on the base classes for Ops, ClientFactory, and Config. 

I took the opportunity to polish the class docstrings and consistently document the class variables and attributes.

### DIFF
--- a/src/orca/services/base/client_factory.py
+++ b/src/orca/services/base/client_factory.py
@@ -2,39 +2,37 @@ from __future__ import annotations
 
 from abc import ABC, abstractmethod
 from functools import cached_property
-from typing import Any, Generic, Type, TypeVar
+from typing import Any, ClassVar, Generic, Type, TypeVar
 
 from pydantic.dataclasses import dataclass
 from typing_extensions import Self
 
-from orca.services.base.config import BaseServiceConfig
+from orca.services.base.config import BaseConfig
 
 ClientClass = TypeVar("ClientClass", bound=Any)
 
-ServiceConfig = TypeVar("ServiceConfig", bound=BaseServiceConfig)
+ConfigClass = TypeVar("ConfigClass", bound=BaseConfig)
 
 
 @dataclass(kw_only=False)
-class BaseClientFactory(ABC, Generic[ClientClass, ServiceConfig]):
-    """Base factory for constructing clients."""
+class BaseClientFactory(ABC, Generic[ClientClass, ConfigClass]):
+    """Base factory for constructing clients.
+
+    Class Variables:
+        connection_env_var: The name of the environment variable whose
+            value is an Airflow connection URI for this service.
+    """
+
+    config_class: ClassVar[Type]
+    client_class: ClassVar[Type]
 
     # Using `__post_init_post_parse__()` to perform steps after validation
     def __post_init_post_parse__(self) -> None:
         """Resolve any attributes using the available methods."""
         self.resolve()
 
-    @property
     @abstractmethod
-    def config_class(self) -> Type[ServiceConfig]:
-        """Service configuration class."""
-
-    @property
-    @abstractmethod
-    def client_class(self) -> Type[ClientClass]:
-        """Service client class."""
-
-    @abstractmethod
-    def update_with_config(self, config: ServiceConfig):
+    def update_with_config(self, config: ConfigClass):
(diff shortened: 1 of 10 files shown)


src/orca/services/base/ops.py
```diff
@@ -1,27 +1,38 @@
...
+ConfigClass = TypeVar("ConfigClass", bound=BaseConfig)
 
 
 @dataclass(kw_only=False, config=ConfigDict(arbitrary_types_allowed=True))
-class BaseOps(ABC, Generic[ServiceConfig, ClientClass]):
-    """Base collection of operations for a service."""
+class BaseOps(ABC, Generic[ConfigClass, ClientClass]):
+    """Base collection of operations for a service.
+
+    N.B. Make sure to override the type hint for the ``client`` class
```
thomasyu888, line 19: TIL: N.B means: nota bene.  It is used to emphasize an important point.

src/orca/services/sevenbridges/hook.py
```diff
@@ -1,77 +1,23 @@
...
 
-    This hook was inspired by the Asana Airflow provider package:
-    https://github.com/apache/airflow/blob/main/airflow/providers/asana/hooks/asana.py
+    Attributes:
+        conn_name_attr: Inherited Airflow attribute (e.g., "sbg_conn_id").
+        default_conn_name: Inherited Airflow attribute (e.g., "sbg_default").
+        conn_type: Inherited Airflow attribute (e.g., "sbg").
+        hook_name: Inherited Airflow attribute (e.g., "SevenBridges").
+
+    Class Variables:
```
thomasyu888, line 14: Good use of class variables here which is different from instance variables which are those Attributes above.

src/orca/services/base/ops.py
```diff
@@ -1,27 +1,38 @@
 from abc import ABC, abstractmethod
-from typing import Any, Generic, TypeVar
+from typing import Any, ClassVar, Generic, Type, TypeVar
 
 from pydantic import ConfigDict
 from pydantic.dataclasses import dataclass
 from typing_extensions import Self
 
-from orca.services.base.config import BaseServiceConfig
+from orca.services.base.config import BaseConfig
```
thomasyu888, line 8: I agree with this change, but can you speak to why you renamed `BaseServiceConfig` to `BaseConfig`?

src/orca/services/base/client_factory.py
```diff
@@ -2,39 +2,37 @@
...
 
 
 @dataclass(kw_only=False)
-class BaseClientFactory(ABC, Generic[ClientClass, ServiceConfig]):
-    """Base factory for constructing clients."""
+class BaseClientFactory(ABC, Generic[ClientClass, ConfigClass]):
+    """Base factory for constructing clients.
+
+    Class Variables:
+        connection_env_var: The name of the environment variable whose
```
thomasyu888, line 22: Are the Class Variables `config_class` and `client_class`?

thomasyu888 (changes requested):
🚀 LGTM: Some small changes

src/orca/services/base/ops.py
BrunoGrandePhD: Mostly for consistency. I didn't like that the base class was `BaseServiceConfig` whereas the implementations would be `SevenBridgesConfig` and `NextflowTowerConfig` (_i.e. without `Service`, because it's redundant and too long). I prefer `Base` being the only part that is changed when inheriting from a base class.

src/orca/services/base/ops.py
BrunoGrandePhD: I didn't realize that was the spelt out version. I read it in Italian in my mind, which sounds nice. I'm going to read it that way going forward (instead of just thinking of the letters N and B).

src/orca/services/base/client_factory.py
BrunoGrandePhD: This is addressed in #11, which has several docstring-related improvements. I did those after making several code changes, which made it hard to distribute those docstring improvements to the individual PRs.

thomasyu888 (approved):

BWMac (approved):

PR #11, opened 2023-03-01
[ORCA-154] Migrate config-related code to `BaseConfig` and improve docstrings

BrunoGrandePhD:
When I tried implementing the Nextflow Tower service, I noticed that several ways to improve the existing base classes, mostly by moving all config-related functionality into `BaseConfig`. This improves the code's adherence to the single-responsibility principle (SRP). This also paves the way to adding support for configuration files to `BaseConfig` without having to change anything about the other base classes. That's a sign that SRP is being followed. 

Unfortunately, I started refactoring on top of #10, so I can't easily merge these changes into #8, where `BaseConfig` is first introduced. I'll leave a comment on #8 about this PR.

### DIFF
--- a/src/orca/services/base/client_factory.py
+++ b/src/orca/services/base/client_factory.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 from abc import ABC, abstractmethod
 from functools import cached_property
-from typing import Any, Generic, TypeVar
+from typing import Any, ClassVar, Generic, Type, TypeVar
 
 from pydantic.dataclasses import dataclass
 
@@ -18,15 +18,23 @@ ConfigClass = TypeVar("ConfigClass", bound=BaseConfig)
 class BaseClientFactory(ABC, Generic[ClientClass, ConfigClass]):
     """Base factory for constructing clients.
 
+    Usage Instructions:
+        1) Create a class that subclasses this base class.
+        2) Decorate this class with ``@dataclass`` as this one does.
+        3) Provide values to all class variables (defined below).
+        4) Provide implementations for all abstract methods.
+        5) Update the type hints for attributes and class variables.
+
     Attributes:
-        config: Configuration relevant to this service.
+        config: Configuration object for this service.
 
     Class Variables:
-        connection_env_var: The name of the environment variable whose
-            value is an Airflow connection URI for this service.
+        client_class: The client class for this service.
     """
 
     config: ConfigClass
+
+    client_class: ClassVar[Type]
 
     @abstractmethod
     def create_client(self) -> ClientClass:
@@ -54,7 +62,7 @@ class BaseClientFactory(ABC, Generic[ClientClass, ConfigClass]):
         is made but indicates a problem.
 
         Args:
-            An authenticated instance of the client for this service.
+            client: An authenticated client for this service.
         """
 
     @classmethod
@@ -62,7 +70,7 @@ class BaseClientFactory(ABC, Generic[ClientClass, ConfigClass]):
         """Test the client with an authenticated request.
 
         Args:
-            An authenticated instance of the client for this service.
+            client: An authenticated client for this service.
 
         Raises:
             ClientRequestError: If an error occured while making a request.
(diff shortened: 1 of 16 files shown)


src/orca/services/base/client_factory.py
```diff
@@ -18,87 +18,71 @@
...
+        That is taken care of by the ``test_client()`` method.
+        That said, this method can raise an error if a response
+        is made but indicates a problem.
 
         Args:
-            config: Arguments relevant to this service.
-
-        Returns:
-            An instantiated client factory.
+            client: An authenticated client for this service.
```
thomasyu888: Nit: what is the difference between this and `An authenticated client object.`?

src/orca/services/base/hook.py
```diff
@@ -25,22 +25,26 @@ class BaseOrcaHook(BaseHook, Generic[OpsClass, ClientClass]):
...
+
+    Class Variables:
         conn_name_attr: Inherited Airflow attribute (e.g., "sbg_conn_id").
         default_conn_name: Inherited Airflow attribute (e.g., "sbg_default").
         conn_type: Inherited Airflow attribute (e.g., "sbg").
         hook_name: Inherited Airflow attribute (e.g., "SevenBridges").
-
-    Class Variables:
         ops_class: The Ops class for this service.
+        config_class: The configuration class for this service.
```
thomasyu888, line 38: Makes sense to add this as a class attribute.

src/orca/services/sevenbridges/client_factory.py
```diff
@@ -35,85 +24,55 @@ class SevenBridgesClientFactory(BaseClientFactory):
...
-            addendum = f"API ({self.api_endpoint}) is not among {valid_api_endpoints}."
-            raise ClientAttrError(common_message + addendum)
-
-    def prepare_client_kwargs(self) -> dict[str, Any]:
-        """Prepare client keyword arguments.
+            ConfigError: If the configuration is invalid.
 
         Returns:
-            Dictionary of keyword arguments.
+            An authenticated client object.
```
thomasyu888, line 48: Nit: Same thing here about referring to authenticated client objects.

src/orca/services/sevenbridges/config.py
```diff
@@ -21,44 +30,58 @@ class variable.
...
 
     connection_env_var = "SEVENBRIDGES_CONNECTION_URI"
 
-    valid_api_endpoints: ClassVar[set[str]] = {
-        "https://api.sbgenomics.com/v2",
-        "https://cgc-api.sbgenomics.com/v2",
-        "https://cavatica-api.sbgenomics.com/v2",
-    }
+    @validator("api_endpoint")
+    def validate_api_endpoint(cls, value: str):
```
thomasyu888, line 49: I can envision this function being helpful in `BaseConfig`, but I am ok with this here.

thomasyu888 (approved):
🚀 Great migration of code - just minor comments.  Heres the code I ran and it worked

```
from orca.services.sevenbridges.client_factory import (
    SevenBridgesConfig,
)

from orca.services.sevenbridges.ops import (
    SevenBridgesOps,
)
auth_token = ""
config = SevenBridgesConfig(
    api_endpoint="https://cavatica-api.sbgenomics.com/v2",
    auth_token=auth_token,
    project="orca-service/test-project"
)
## No longer need to define this because SevenBridgesClientFactory
## is a class attribute of SevenBridgesOps
# factory = SevenBridgesClientFactory(
#     config=config
# )
# client = factory.get_client()
ops = SevenBridgesOps(
    config=config
)

ops.client.tasks.query(project="orca-service/test-project")
```

BWMac (approved):

src/orca/services/sevenbridges/config.py
BrunoGrandePhD: Not all services might have a restricted set of API endpoints like SevenBridges. If it's a recurring pattern, we could create a mixin with the Pydantic validation function.

PR #13, opened 2023-03-01
[ORCA-155] Implement basic Nextflow Tower service

BrunoGrandePhD:
I wanted to use the new base classes to implement a new service, so I chose Nextflow Tower. This implementation is very basic, but it serves to demonstrate the utility of the base classes. 

I realized that I can probably parametrize the tests as well for different services to avoid duplication. I'll look into that next. 

- [x] Implement tests for Nextflow Tower

### DIFF
--- a/src/orca/services/nextflowtower/client.py
+++ b/src/orca/services/nextflowtower/client.py
@@ -1,0 +1,128 @@
+from typing import Any
+
+import requests
+from pydantic.dataclasses import dataclass
+
+
+@dataclass(kw_only=False)
+class NextflowTowerClient:
+    """Simple Python client for making requests to Nextflow Tower.
+
+    Attributes:
+        api_endpoint: API endpoint for a Nextflow Tower platform.
+        auth_token: An authentication token for the platform specified
+            by the ``api_endpoints`` value.
+    """
+
+    auth_token: str
+    api_endpoint: str
+
+    def update_kwarg(
+        self, kwargs: dict[str, Any], key1: str, key2: str, default: Any
+    ) -> None:
+        """Ensure a default value for a nested key in kwargs.
+
+        Args:
+            kwargs: Keyword arguments
+            key1: Top-level key.
+            key2: Nested key.
+            default: Default value.
+        """
+        kwargs.setdefault(key1, dict())
+        if not isinstance(kwargs[key1], dict):
+            message = f"The '{key1}' keyword argument must be a dictionary."
+            raise ValueError(message)
+        kwargs[key1].setdefault(key2, default)
+        if not isinstance(kwargs[key1][key2], type(default)):
+            message = (
+                f"The value for kwargs['{key1}']['{key2}'] ({kwargs[key1][key2]}) "
+                f"is not the expected type ({type(default)})."
+            )
+            raise ValueError(message)
+
+    def request(self, method: str, path: str, **kwargs) -> requests.Response:
+        """Make an authenticated HTTP request.
+
+        Args:
+            method: An HTTP method (GET, PUT, POST, or DELETE).
+            path: The API path with the parameters filled in.
+            **kwargs: Additional named arguments passed through to
+                requests.request().
+
+        Returns:
+            The raw Response object to allow for special handling
+        """
+        valid_methods = {"GET", "PUT", "POST", "DELETE"}
+        if method not in valid_methods:
+            message = f"Method ({method}) not among valid options ({valid_methods})."
+            raise ValueError(message)
+
+        url = f"{self.api_endpoint}/{path}"
+
+        auth_header = f"Bearer {self.auth_token}"
+        self.update_kwarg(kwargs, "headers", "Authorization", auth_header)
+
+        return requests.request(method, url, **kwargs)
+
+    def request_json(self, method: str, path: str, **kwargs) -> dict[str, Any]:
+        """Make an authenticated HTTP request and parse the JSON response.
+
+        See ``TowerClient.request`` for argument definitions.
+
+        Raises:
+            HTTPError: If something went wrong with the request.
+
+        Returns:
+            A dictionary from deserializing the JSON response.
+        """
+        response = self.request(method, path, **kwargs)
+        response.raise_for_status()
+        return response.json()
+
+    def request_paged(self, method: str, path: str, **kwargs) -> list[dict[str, Any]]:
+        """Iterate through pages of results for a given request.
+
+        See ``TowerClient.request`` for argument definitions.
+
+        Raises:
+            ValueError: If the 'params' keyword argument
+                isn't a dictionary
+
+        Returns:
+            The cumulative list of items from all pages.
+        """
+        self.update_kwarg(kwargs, "params", "max", 50)
+        self.update_kwarg(kwargs, "params", "offset", 0)
+
+        num_items = 0
+        total_size = 1  # Artificial value for initiating the while-loop
+
+        all_items = list()
+        while num_items < total_size:
+            kwargs["params"]["offset"] = num_items
+            json = self.request_json(method, path, **kwargs)
+
+            if "totalSize" not in json:
+                message = f"'totalSize' not in response JSON ({json}) as expected."
+                raise requests.exceptions.HTTPError(message)
+            total_size = json.pop("totalSize")
+
+            if len(json) != 1:
+                message = f"Expected one other key aside from 'totalSize' ({json})."
+                raise requests.exceptions.HTTPError(message)
+            _, items = json.popitem()
+
+            num_items += len(items)
+            all_items.extend(items)
+
+        return all_items
+
+    # TODO: Consider creating a `client` submodule folder to organize methods
+    def get_user_info(self) -> dict[str, Any]:
+        """Describe current user.
+
+        Returns:
+            Current user
+        """
+        path = "/user-info"
+        return self.request_json("get", path)
--- a/src/orca/services/nextflowtower/client.py
+++ b/src/orca/services/nextflowtower/client.py
@@ -2,6 +2,7 @@ from typing import Any
 
 import requests
 from pydantic.dataclasses import dataclass
+from requests.exceptions import HTTPError
 
 
 @dataclass(kw_only=False)
@@ -27,6 +28,11 @@ class NextflowTowerClient:
             key1: Top-level key.
             key2: Nested key.
             default: Default value.
+
+        Raises:
+            ValueError: If 'params' isn't a dictionary.
+            ValueError: If the existing value under the provided keys
+                does not match the type of the default value.
         """
         kwargs.setdefault(key1, dict())
         if not isinstance(kwargs[key1], dict):
@@ -49,6 +55,9 @@ class NextflowTowerClient:
             **kwargs: Additional named arguments passed through to
                 requests.request().
 
+        Raises:
+            ValueError: If the provided method isn't valid.
+
         Returns:
             The raw Response object to allow for special handling
         """
@@ -65,7 +74,7 @@ class NextflowTowerClient:
         return requests.request(method, url, **kwargs)
 
     def request_json(self, method: str, path: str, **kwargs) -> dict[str, Any]:
-        """Make an authenticated HTTP request and parse the JSON response.
+        """Make an auth'ed HTTP request and parse the JSON response.
 
         See ``TowerClient.request`` for argument definitions.
 
@@ -85,8 +94,8 @@ class NextflowTowerClient:
         See ``TowerClient.request`` for argument definitions.
 
         Raises:
-            ValueError: If the 'params' keyword argument
-                isn't a dictionary
+            HTTPError: If the response doesn't match the expectation
+                for a paged endpoint.
 
         Returns:
             The cumulative list of items from all pages.
@@ -104,18 +113,28 @@ class NextflowTowerClient:
 
             if "totalSize" not in json:
                 message = f"'totalSize' not in response JSON ({json}) as expected."
-                raise requests.exceptions.HTTPError(message)
+                raise HTTPError(message)
             total_size = json.pop("totalSize")
 
             if len(json) != 1:
                 message = f"Expected one other key aside from 'totalSize' ({json})."
-                raise requests.exceptions.HTTPError(message)
+                raise HTTPError(message)
             _, items = json.popitem()
 
             num_items += len(items)
             all_items.extend(items)
 
         return all_items
+
+    def get(self, path: str, **kwargs) -> dict[str, Any]:
+        """Send an auth'ed GET request and parse the JSON response.
+
+        See ``TowerClient.request`` for argument definitions.
+
+        Returns:
+            A dictionary from deserializing the JSON response.
+        """
+        return self.request_json("GET", path, **kwargs)
 
     # TODO: Consider creating a `client` submodule folder to organize methods
     def get_user_info(self) -> dict[str, Any]:
@@ -125,4 +144,4 @@ class NextflowTowerClient:
             Current user
         """
         path = "/user-info"
-        return self.request_json("get", path)
+        return self.get(path)
--- a/src/orca/services/nextflowtower/client.py
+++ b/src/orca/services/nextflowtower/client.py
@@ -19,8 +19,9 @@ class NextflowTowerClient:
     auth_token: str
     api_endpoint: str
 
+    @staticmethod
     def update_kwarg(
-        self, kwargs: dict[str, Any], key1: str, key2: str, default: Any
+        kwargs: dict[str, Any], key1: str, key2: str, default: Any
     ) -> None:
         """Ensure a default value for a nested key in kwargs.
 
@@ -62,11 +63,6 @@ class NextflowTowerClient:
         Returns:
             The raw Response object to allow for special handling
         """
-        valid_methods = {"GET", "PUT", "POST", "DELETE"}
-        if method not in valid_methods:
-            message = f"Method ({method}) not among valid options ({valid_methods})."
-            raise ValueError(message)
-
         url = f"{self.api_endpoint}/{path}"
 
         auth_header = f"Bearer {self.auth_token}"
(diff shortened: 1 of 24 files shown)


BrunoGrandePhD:
I'll mark this as ready for review to get feedback on the existing PR, but I plan on adding tests before merging.

src/orca/services/nextflowtower/client.py
```diff
@@ -0,0 +1,128 @@
...
+    """Simple Python client for making requests to Nextflow Tower.
+
+    Attributes:
+        api_endpoint: API endpoint for a Nextflow Tower platform.
+        auth_token: An authentication token for the platform specified
+            by the ``api_endpoints`` value.
+    """
+
+    auth_token: str
+    api_endpoint: str
```
BWMac, line 20: In `sagetasks` we had a set list of API endpoints in a dictionary similar to the `SevenBridges` implementation in this repo. Is the reason to not include that here to allow for non-sage use cases?

src/orca/services/nextflowtower/client.py
```diff
@@ -0,0 +1,128 @@
...
+            raise ValueError(message)
+        kwargs[key1].setdefault(key2, default)
+        if not isinstance(kwargs[key1][key2], type(default)):
+            message = (
+                f"The value for kwargs['{key1}']['{key2}'] ({kwargs[key1][key2]}) "
+                f"is not the expected type ({type(default)})."
+            )
+            raise ValueError(message)
+
+    def request(self, method: str, path: str, **kwargs) -> requests.Response:
```
BWMac, line 51: I like having `request` as an option because it gives you full access to the tower API functionality, but just want to confirm that more convenient methods like `launch_workflow` from `sagetasks` will be a part of future work

src/orca/services/nextflowtower/client.py
BrunoGrandePhD: In the case of SevenBridges, there are no other possible options for `api_endpoint`. On the other hand, there are multiple deployments of Nextflow Tower. So, you are correct in saying that this decision is to allow for non-Sage use cases.

src/orca/services/nextflowtower/client.py
BrunoGrandePhD: That's correct! We can add the convenience methods like `launch_workflow()` as we add the correponding function in `NextflowTowerOps`.

src/orca/services/nextflowtower/client.py
```diff
@@ -0,0 +1,128 @@
...
+        """
+        response = self.request(method, path, **kwargs)
+        response.raise_for_status()
+        return response.json()
+
+    def request_paged(self, method: str, path: str, **kwargs) -> list[dict[str, Any]]:
+        """Iterate through pages of results for a given request.
+
+        See ``TowerClient.request`` for argument definitions.
+
```
thomasyu888, line 92: Nit: Add documentation all the way through about **kwargs being `requests.requests(**kwargs)`

src/orca/services/nextflowtower/client.py
```diff
@@ -0,0 +1,128 @@
...
+        Raises:
+            ValueError: If the 'params' keyword argument
+                isn't a dictionary
+
+        Returns:
+            The cumulative list of items from all pages.
+        """
+        self.update_kwarg(kwargs, "params", "max", 50)
```
thomasyu888: I understand the reason for these two parameters and actually like the way you've implemented it, but what is your reason for not including `max` and `offset` as an argument (`def request_paged(...., max=50...)`) and as a kwarg?

src/orca/services/nextflowtower/client.py
```diff
@@ -0,0 +1,128 @@
...
+        response.raise_for_status()
+        return response.json()
+
+    def request_paged(self, method: str, path: str, **kwargs) -> list[dict[str, Any]]:
+        """Iterate through pages of results for a given request.
+
+        See ``TowerClient.request`` for argument definitions.
+
+        Raises:
+            ValueError: If the 'params' keyword argument
```
thomasyu888: Nit: Raise HTTPError?

src/orca/services/nextflowtower/config.py
```diff
@@ -0,0 +1,55 @@
...
+        workspace: A Nextflow Tower workspace ID.
+
+    Class Variables:
+        connection_env_var: The name of the environment variable whose
+            value is an Airflow connection URI for this service.
+    """
+
+    api_endpoint: Optional[str] = None
+    auth_token: Optional[str] = None
+    workspace: Optional[int] = None
```
thomasyu888: Should this be `Optional[str]`?

thomasyu888 (changes requested):
Awesome work, but the below failed for me.

```
from orca.services.nextflowtower.client_factory import (
    NextflowTowerConfig,
)

from orca.services.nextflowtower.ops import (
    NextflowTowerOps,
)
auth_token = ""

# I modified the type in Config as well to support this workspace.
config = NextflowTowerConfig(
    api_endpoint="https://tower.sagebionetworks.org/api",
    auth_token=auth_token,
    workspace="genie-bpc-project"
)

ops = NextflowTowerOps(
    config=config
)

ops.client
```

This caused this error:

```
Traceback (most recent call last):
  File "/Users/tyu/sage/py-orca/src/orca/services/base/client_factory.py", line 79, in test_client
    cls.test_client_request(client)
  File "/Users/tyu/sage/py-orca/src/orca/services/nextflowtower/client_factory.py", line 62, in test_client_request
    client.get_user_info()
  File "/Users/tyu/sage/py-orca/src/orca/services/nextflowtower/client.py", line 128, in get_user_info
    return self.request_json("get", path)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/tyu/sage/py-orca/src/orca/services/nextflowtower/client.py", line 78, in request_json
    response = self.request(method, path, **kwargs)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/tyu/sage/py-orca/src/orca/services/nextflowtower/client.py", line 58, in request
    raise ValueError(message)
ValueError: Method (get) not among valid options ({'PUT', 'GET', 'POST', 'DELETE'}).

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/Cellar/python@3.11/3.11.1/Frameworks/Python.framework/Versions/3.11/lib/python3.11/functools.py", line 1001, in __get__
    val = self.func(instance)
          ^^^^^^^^^^^^^^^^^^^
  File "/Users/tyu/sage/py-orca/src/orca/services/base/ops.py", line 39, in client
    return factory.get_client(test=True)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/tyu/sage/py-orca/src/orca/services/base/client_factory.py", line 100, in get_client
    self.test_client(client)
  File "/Users/tyu/sage/py-orca/src/orca/services/base/client_factory.py", line 82, in test_client
    raise ClientRequestError(message) from error
orca.errors.ClientRequestError: Authenticated test request failed using the client.
>>> ops.client
Traceback (most recent call last):
  File "/Users/tyu/sage/py-orca/src/orca/services/base/client_factory.py", line 79, in test_client
    cls.test_client_request(client)
  File "/Users/tyu/sage/py-orca/src/orca/services/nextflowtower/client_factory.py", line 62, in test_client_request
    client.get_user_info()
  File "/Users/tyu/sage/py-orca/src/orca/services/nextflowtower/client.py", line 128, in get_user_info
    return self.request_json("get", path)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/tyu/sage/py-orca/src/orca/services/nextflowtower/client.py", line 78, in request_json
    response = self.request(method, path, **kwargs)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/tyu/sage/py-orca/src/orca/services/nextflowtower/client.py", line 58, in request
    raise ValueError(message)
ValueError: Method (get) not among valid options ({'PUT', 'GET', 'POST', 'DELETE'}).

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/Cellar/python@3.11/3.11.1/Frameworks/Python.framework/Versions/3.11/lib/python3.11/functools.py", line 1001, in __get__
    val = self.func(instance)
          ^^^^^^^^^^^^^^^^^^^
  File "/Users/tyu/sage/py-orca/src/orca/services/base/ops.py", line 39, in client
    return factory.get_client(test=True)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/tyu/sage/py-orca/src/orca/services/base/client_factory.py", line 100, in get_client
    self.test_client(client)
  File "/Users/tyu/sage/py-orca/src/orca/services/base/client_factory.py", line 82, in test_client
    raise ClientRequestError(message) from error
orca.errors.ClientRequestError: Authenticated test request failed using the client
```

src/orca/services/nextflowtower/client.py
```diff
@@ -0,0 +1,128 @@
...
+
+    # TODO: Consider creating a `client` submodule folder to organize methods
+    def get_user_info(self) -> dict[str, Any]:
+        """Describe current user.
+
+        Returns:
+            Current user
+        """
+        path = "/user-info"
+        return self.request_json("get", path)
```
thomasyu888: I think this needs to be uppercase? "GET" (Or see my previous comment)

src/orca/services/nextflowtower/client.py
```diff
@@ -0,0 +1,128 @@
...
+        Args:
+            method: An HTTP method (GET, PUT, POST, or DELETE).
+            path: The API path with the parameters filled in.
+            **kwargs: Additional named arguments passed through to
+                requests.request().
+
+        Returns:
+            The raw Response object to allow for special handling
+        """
+        valid_methods = {"GET", "PUT", "POST", "DELETE"}
```
thomasyu888: Nit: you could uppercase the `method` here to allow for lower case methods as parameters in functions.

src/orca/services/nextflowtower/client.py
BrunoGrandePhD: The docstring already states:

>  See ``TowerClient.request`` for argument definitions.

And `TowerClient.request` defines the following:

> **kwargs: Additional named arguments passed through to requests.request().

Is there something that you want to add? To clarify, my intent here is to avoid duplication of documentation.

src/orca/services/nextflowtower/client.py
BrunoGrandePhD: For consistency with the other `request` methods and for simplicity. For example, if I add a `max` parameter, I now need to handle (and document) the case where a user specifies `max` in the parameter and in the `params` dictionary. The user already has a way to provide this value, so I don't see the need to add another way, especially since I don't expect users to edit this value very often.

src/orca/services/nextflowtower/client.py
thomasyu888: 100% just glossed over that.  Thanks for pointing that out!

src/orca/services/nextflowtower/client.py
BrunoGrandePhD: Good catch! I reviewed all of the docstrings and made sure that the `Raises` section is consistent.

src/orca/services/nextflowtower/config.py
BrunoGrandePhD: That's the type of the organization and workspace IDs according to the Tower OpenAPI [specification](https://api.tower.nf/openapi#put-/orgs/-orgId-/workspaces/-workspaceId-). 

image

src/orca/services/nextflowtower/client.py
BrunoGrandePhD: Yeah, that's a bug! I added a convenience `get()` method to minimize the risk for this error.

src/orca/services/nextflowtower/client.py
BrunoGrandePhD: I considered that, but I prefer the `get()` convenience method, which is more elegant:

```python
return self.request_json("GET", path)
# versus
return self.get(path)
```

src/orca/services/nextflowtower/client.py
thomasyu888: +1

src/orca/services/nextflowtower/config.py
BrunoGrandePhD: I just noticed your comment [here](https://github.com/Sage-Bionetworks-Workflows/py-orca/pull/13#pullrequestreview-1331380427). I realize the confusion. While the docstring specifies that this value should be the numerical workspace ID, the attribute name (no `_id`) doesn't imply that. I agree that it's convenient to specify workspace names. The Tower CLI supports this with this format: `<org-name>/<workspace-name>` (similar to project names in SevenBridges). I'll add support for both methods.

src/orca/services/nextflowtower/client.py
thomasyu888: 👍🏼

BrunoGrandePhD:
@thomasyu888: Your feedback was very valuable. I've updated the config to expect the fully-qualified workspace name (_e.g._ `sage-bionetworks/genie-bpc-project`), which are more user-friendly than workspace IDs. I have added a cached property to the Ops class to retrieve the workspace ID once when requested since most workspace-related requests require the ID and not the name. 

The following code should now work for you: 

```python
from orca.services.nextflowtower import NextflowTowerConfig, NextflowTowerOps

auth_token = "REDACTED"

config = NextflowTowerConfig(
    api_endpoint="https://tower.sagebionetworks.org/api",
    auth_token=auth_token,
    workspace="sage-bionetworks/genie-bpc-project",
)

ops = NextflowTowerOps(config)
ops.client
```

thomasyu888 (commented):
Excellent update @BrunoGrandePhD .  Just a question, when I run this:

```
from orca.services.nextflowtower import (
    NextflowTowerConfig,
    NextflowTowerOps
)
auth_token = ""
config = NextflowTowerConfig(
    api_endpoint="https://tower.sagebionetworks.org/api",
    auth_token=auth_token,
    workspace="sage-bionetworks/genie-bpc-project"
)

# response = ops.client.request("GET", "/pipelines")
# response.content
# response = ops.client.request("GET", "/workflow/random-name")
# response.content

# This shows 0 pipelines even though I see 3 pipelines in the UI
ops.client.get("/pipelines")
# This works
ops.client.get("/workflow/random-name")
# This works
ops.workspace_id
# 5355285966491
```

Is this a permission issue with the access token that I don't see any pipelines even though I see them in the UI?

BrunoGrandePhD:
Good catch! That's happening because `NextflowTowerClient` will not automatically populate the `workspaceId` query parameter. This is something that needs to be done explicitly (see below). Otherwise, you are presumably pulling the list of pipelines in your user workspace, which we disabled on the Sage Tower instance. 

My plan is to make the `NextflowTowerOps` class workspace-aware similar to how the `SevenBridgesOps` class is project-aware. I opted against making the `NextflowTowerClient` workspace-aware for consistency with the SevenBridges client. That said, we can revisit this. Ideally, most common operations will be implemented in `NextflowTowerOps`, obviating the need to explicitly set the workspace. 

```python
ops.client.get("/pipelines", params={"workspaceId": ops.workspace_id})
```

thomasyu888 (commented):
🦖 Thanks @BrunoGrandePhD !  That makes sense and your command worked!

```
ops.client.get("/pipelines", params={"workspaceId": ops.workspace_id})
```

We should have a short meeting about adding tickets to this EPIC [Nextflow Tower Hooks EPIC](https://sagebionetworks.jira.com/browse/ORCA-144) .  This way we can start distributing some of the work since you've put together this beautiful framework.

Before I approve - I noticed you have a task to add more tests - is that still true?  Or... do you think we should add them in as we add in functions in `ops`?

tests/services/nextflowtower/responses.py
```diff
@@ -0,0 +1,65 @@
...
+To generate these, you can visit the OpenAPI page linked below and
+make example requests after setting your authentication token at the
+top of the page. You'll have to replace JSON literals ('null', 'false',
+'true') with Python equivalents ('None', 'False', 'True'). Take care
+of redacting any values that shouldn't be public.
+
+https://tower-dev.sagebionetworks.org/api/openapi
+"""
+
+get_user_info = {
```
thomasyu888, line 12: What do you think about these making these as fixtures and yielding them to act as setup and teardown?

src/orca/services/nextflowtower/config.py
```diff
@@ -0,0 +1,81 @@
...
+            value is an Airflow connection URI for this service.
+    """
+
+    api_endpoint: Optional[str] = None
+    auth_token: Optional[str] = None
+    workspace: Optional[str] = None
+
+    connection_env_var = "NEXTFLOWTOWER_CONNECTION_URI"
+
+    @validator("workspace")
```
BWMac, line 37: TIL the validator decorator. Very cool.

BWMac (approved):
LGTM. I used the following code and everything worked as expected:
```
#import
from orca.services.nextflowtower.client_factory import NextflowTowerClientFactory
from orca.services.nextflowtower.config import NextflowTowerConfig
from orca.services.nextflowtower.ops import NextflowTowerOps

#confiugration variables
auth_token = os.environ.get("TOWER_ACCESS_TOKEN")
api_endpoint = "https://tower.sagebionetworks.org/api"
workspace = "Sage-Bionetworks/example-project"

#set config

#create factory
factory = NextflowTowerClientFactory(config=config)

#create client
client = factory.create_client()

#test client functionality
info = client.get_user_info()
client.get_user_workspaces_and_orgs(info["id"])
client.list_user_workspaces()

#create ops

#test ops properties
ops.workspace_id
ops.workspace
```

thomasyu888 (approved):
💯 awesome! Just a minor comment, but going to pre-approve.

tests/services/nextflowtower/responses.py
BrunoGrandePhD: I'm not a fan of having to include another fixture in every test function that needs to use a response, but if you're worried about ensuring that tests are independent by preventing the modification of mutable shared dictionaries, then there's not much choice. I addressed this here: https://github.com/Sage-Bionetworks-Workflows/py-orca/pull/13/commits/5e3a3bee36c7d258f1e6b1d5e97301f0579654e1

PR #14, opened 2023-04-24
[ORCA-155] Adding tests to base tower service class

rxu17:
**Purpose:** This PR (unfinished) creates tests for the base tower service class so we can test the base class whilest reducing redundancy in tests for the tower specific services for only the following classes:

- config
- ops
- client_factory

This PR is still a work in progress, and some of the tests need troubleshooting (mainly in `tests/services/base/test_config.py` due to some of the mocking used). Haven't been able to put much time in this PR for the past sprint (since this PR was created)

### DIFF
--- a/tests/services/base/conftest.py
+++ b/tests/services/base/conftest.py
@@ -1,0 +1,98 @@
+import pytest
+from airflow.models.connection import Connection
+
+from orca.services.nextflowtower import NextflowTowerConfig, NextflowTowerHook
+from orca.services.sevenbridges import SevenBridgesConfig, SevenBridgesHook
+
+SUBCLASSES = {
+    SevenBridgesHook: {
+        "api_endpoint": "https://api.sbgenomics.com/v2",
+        "auth_token": "foo",
+        "project": "bgrande/sandbox",
+    },
+    NextflowTowerHook: {
+        "api_endpoint": "https://tower-dev.sagebionetworks.org/api",
+        "auth_token": "foo",
+        "workspace": "sage-bionetworks/example-project",
+    },
+}
+
+SUBCLASS_CONFIG_QUERY_KEYS = {
+    SevenBridgesConfig: "project",
+    NextflowTowerConfig: "workspace",
+}
+
+
+@pytest.fixture(
+    params=SUBCLASSES.items(), ids=[subclass.__name__ for subclass in SUBCLASSES]
+)
+def ops(request):
+    hook_cls, params = request.param
+    config = hook_cls.config_class(**params)
+    ops = hook_cls.ops_class(config)
+    yield ops
+
+
+@pytest.fixture()
+def client_factory(ops):
+    yield ops.client_factory_class(ops.config)
+
+
+@pytest.fixture(params=SUBCLASSES.keys())
+def blank_client_factory(request):
+    hook_cls = request.param
+    config = hook_cls.config_class()
+    ops = hook_cls.ops_class(config)
+    yield ops.client_factory_class(config)
+
+
+@pytest.fixture()
+def client_class(ops):
+    yield ops.client_factory_class(ops.config).client_class
+
+
+@pytest.fixture()
+def config(ops):
+    yield ops.config
+
+
+@pytest.fixture
+def connection_uri(config):
+    bare_url = config.api_endpoint.replace("https://", "")
+    host, schema = bare_url.rstrip("/").rsplit("/", maxsplit=1)
+    token = config.auth_token
+    query_key = SUBCLASS_CONFIG_QUERY_KEYS[config.__class__]
+    query_val = getattr(config, query_key)
+    yield f"sbg://:{token}@{host}/{schema}/?{query_key}={query_val}"
+
+
+@pytest.fixture
+def connection(connection_uri):
+    yield Connection(uri=connection_uri)
+
+
+@pytest.fixture(params=SUBCLASSES.keys())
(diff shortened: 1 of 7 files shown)


BrunoGrandePhD:
I refactored the `conftest.py` to be simpler. There's still some uncovered parts of the Nextflow Tower codebase, but those bits are Tower-specific. I'll update #13 with some additional tests to close these gaps. This PR is intended to focus on the base tests. 

```
Name                                                Stmts   Miss Branch BrPart  Cover   Missing
-----------------------------------------------------------------------------------------------
src/orca/__init__.py                                    7      0      0      0   100%
src/orca/airflow/__init__.py                            2      0      0      0   100%
src/orca/airflow/provider_info.py                       7      0      0      0   100%
src/orca/errors.py                                      4      0      0      0   100%
src/orca/services/__init__.py                           0      0      0      0   100%
src/orca/services/base/__init__.py                      5      0      0      0   100%
src/orca/services/base/client_factory.py               33      0     16      0   100%
src/orca/services/base/config.py                       38      0     20      0   100%
src/orca/services/base/hook.py                         35      0      6      0   100%
src/orca/services/base/ops.py                          14      0      4      0   100%
src/orca/services/nextflowtower/__init__.py             5      0      0      0   100%
src/orca/services/nextflowtower/client.py              77     45     22      3    37%   40-41, 44-48, 67-68, 90, 104-128, 152-155, 166-172, 181-190
src/orca/services/nextflowtower/client_factory.py      22      0      6      0   100%
src/orca/services/nextflowtower/config.py              30      3     12      2    88%   55-57, 71->76
src/orca/services/nextflowtower/hook.py                10      0      0      0   100%
src/orca/services/nextflowtower/ops.py                 25     11     12      0    54%   29-35, 40-43
src/orca/services/sevenbridges/__init__.py              5      0      0      0   100%
src/orca/services/sevenbridges/client_factory.py       27      0      6      0   100%
src/orca/services/sevenbridges/config.py               29      0     10      0   100%
src/orca/services/sevenbridges/hook.py                 10      0      0      0   100%
src/orca/services/sevenbridges/ops.py                  51      0     16      0   100%
-----------------------------------------------------------------------------------------------
TOTAL                                                 436     59    130      5    85%
```

tests/services/base/conftest.py
```diff
@@ -0,0 +1,76 @@
...
+    nextflowtower: "tower://:foo@api.tower.nf/?workspace=bar/baz",
+    sevenbridges: "sbg://:foo@api.sbgenomics.com/v2/?project=bar/baz",
+}
+
+
+def get_service_name(service_module: ModuleType) -> str:
+    return service_module.__name__.split(".")[-1]
+
+
+@pytest.fixture(params=SERVICES, ids=get_service_name)
```
thomasyu888, line 19: TIL: you can pass in a function in the `ids`

This is an elegant solution to run the tests across multiple service. Nice!

Nit: We may want to add in the CONTRIBUTING guide to add to the `SERVICES` dict when new services are added to add to the test

tests/services/base/conftest.py
```diff
@@ -0,0 +1,76 @@
...
+def service(request, patch_os_environ):
```
thomasyu888, line 20: Am I missing in the documentation somewhere that `patch__os_environ` is a provided fixture?  Or is this to support the mocking of the os.environ in the config tests?

tests/services/base/test_client_factory.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+def test_that_client_factory_can_generate_client(client_factory):
+    client = client_factory.create_client()
+    assert isinstance(client, client_factory.client_class)
+
+
+def test_for_an_error_when_testing_the_client_without_auth(client_factory):
+    client = client_factory.create_client()
+    with pytest.raises(ClientRequestError):
```
thomasyu888, line 24: Nit: Did we want to do light matching of the error message? This extends to other pytest.raises, but what i like about this is we can update the error message more easily. 

So maybe not...

thomasyu888 (approved):
💯 LGTM!  Thanks @rxu17 for your efforts on this and @BrunoGrandePhD for bringing this home.  Just some minor comments / questions, but going to pre-approve.

tests/services/base/conftest.py
BrunoGrandePhD: Good idea! Instead of documenting this, I added a test to make sure that all services are listed in that dictionary (`test_that_all_implemented_services_are_being_tested`). This way, it can be enforced automatically.

tests/services/base/conftest.py
BrunoGrandePhD: This is a custom fixture that's implemented in the top-level `conftest.py` file. It's empties the environment to make sure that any connection URIs loaded from the `.env` file are ignored. 

```python
@pytest.fixture
def patch_os_environ(mocker):
    yield mocker.patch("os.environ", {})
```

tests/services/base/test_client_factory.py
BrunoGrandePhD: IMHO, I don't think we should test for error messages because that's not formally part of the interface. In the docstrings, we indicate that an exception class will be raised if a certain situation is encountered. That's part of the interface. I think of the error message as more of an implementation detail.

BrunoGrandePhD:
I just realized that I forgot to commit my local changes to this branch before merging, which addressed Tom's comments. I'll push these to #13 instead.

PR #15, opened 2023-04-27
[ORCA-163] Implement `launch_workflow()` and `LaunchInfo` dataclass

BrunoGrandePhD:
I was initially going to port the `launch_workflow()` implementation from sagetasks ([source](https://github.com/Sage-Bionetworks-Workflows/sagetasks/blob/524b5112a2d7e6b083304e0528fa8d7876f8601a/src/sagetasks/nextflowtower/utils.py#L189-L272)), but I decided to refactor it. Its reliance on the `init_launch_workflow_data()` method is clunky. I also didn't like the large number of function parameters. 

To address this, I implemented the `LaunchInfo` dataclass. Its goal is to encapsulate the information related to launching a workflow. It essentially operates like an OpenAPI model. This results in much cleaner client code. This class can eventually be reused for `get_workflow()`. 

I'm marking this PR as a draft because I would like to tackle the following subtasks:

- [x] Implement `get_compute_env()` in the client class
- [x] Implement `get_latest_compute_env()` in the ops class
- [x] Implement higher-level `launch_workflow()` in the ops class
  - This version would use the default values from the compute environment (_e.g._ work directory) when launching a workflow ([example](https://github.com/Sage-Bionetworks-Workflows/sagetasks/blob/524b5112a2d7e6b083304e0528fa8d7876f8601a/src/sagetasks/nextflowtower/utils.py#L156-L159))
  - This version would ideally be idempotent, if possible


src/orca/services/nextflowtower/models.py
```diff
@@ -0,0 +1,85 @@
...
+    @staticmethod
+    def get_current_timestamp():
+        """Generate current timestamp using Tower's UTC format.
+
+        Returns:
+            Current timestamp.
+        """
+        return datetime.now().isoformat()[:-3] + "Z"
+
+    def to_dict(self) -> dict[str, Any]:
```
thomasyu888, line 129: Is this the response people get when they call: `dataclasses.asdict(obj)`?

src/orca/services/nextflowtower/models.py
thomasyu888, line 1: Thanks for doing this - it makes the code a lot cleaner!  Not only can we pass this model around and but it also adds a layer of documentation for the request object without going to the API.

src/orca/services/nextflowtower/models.py
```diff
@@ -2,14 +2,68 @@
...
+@dataclass(kw_only=False)
+class User:
+    """Nextflow Tower user."""
+
+    id: int
+    username: str
+    email: str
+
+    @classmethod
+    def from_response(cls, response: dict[str, Any]) -> Self:
```
thomasyu888: Nit: https://docs.pydantic.dev/usage/models/#helper-functions.  Would this be helpful? (in particular the `parse_obj` function included in `BaseModels`)

That said, that is technically beyond the scope of this PR.

BrunoGrandePhD: That doesn't work for most models because the API responses use camel case and the data classes use snake case. We could make this work if we used camel case for our data classes. Not sure if we want to break convention though.

src/orca/services/nextflowtower/models.py
BrunoGrandePhD: See my comment [here](https://github.com/Sage-Bionetworks-Workflows/py-orca/pull/15#discussion_r1180847295).

src/orca/services/nextflowtower/models.py
thomasyu888: Ya... let's not break the convention. Was just curious if we could reuse something.

src/orca/services/nextflowtower/client.py
```diff
@@ -85,43 +86,36 @@ def request_json(self, method: str, path: str, **kwargs) -> dict[str, Any]:
...
-    #         all_items.extend(items)
+        all_items = list()
+        total_size = float("inf")  # Artificial value for initiating the while-loop
+        while num_items < total_size:
+            kwargs["params"]["offset"] = num_items
+            json = self.request_json(method, path, **kwargs)
+            total_size = json.pop("totalSize")
+            _, items = json.popitem()
+            num_items += len(items)
+            all_items.extend(items)
```
thomasyu888, line 115: What is your rationale of returning a list rather than a generator?

src/orca/services/nextflowtower/client.py
```diff
@@ -131,56 +125,205 @@ def get(self, path: str, **kwargs) -> dict[str, Any]:
...
             A dictionary from deserializing the JSON response.
         """
-        return self.request_json("GET", path, **kwargs)
+        response = self.request_json("GET", path, **kwargs)
+        if "totalSize" in response:
+            total_size = response.pop("totalSize")
+            key, items = response.popitem()
+            kwargs["params"] = kwargs["params"] or dict()
+            kwargs["params"]["offset"] = len(items)
+            remainder = self.request_paged("GET", path, **kwargs)
```
thomasyu888: I appreciate your intention for saving one rest call, but what are your thoughts on the added complexity of the code over the one rest call that you save?

src/orca/services/nextflowtower/client.py
```diff
@@ -131,56 +125,205 @@ def get(self, path: str, **kwargs) -> dict[str, Any]:
...
-    def get_user_info(self) -> dict[str, Any]:
-        """Describe current user.
+        See ``TowerClient.request`` for argument definitions.
 
-        Raises:
-            HTTPError: If the response lacks the expected keys.
+        Returns:
+            A dictionary from deserializing the JSON response.
+        """
+        return self.request_json("POST", path, **kwargs)
```
thomasyu888, line 145: Is there a need to add support for pagination here?

src/orca/services/nextflowtower/client.py
```diff
@@ -131,56 +125,205 @@ def get(self, path: str, **kwargs) -> dict[str, Any]:
...
+        Returns:
+            Label instance.
+        """
+        path = "/labels"
+        params = self.generate_params(workspace_id)
+        payload = {"name": name, "resource": False}
+        response = self.post(path, params=params, json=payload)
+        return models.Label.from_response(response)
+
+    def list_labels(self, workspace_id: Optional[int] = None) -> list[models.Label]:
```
thomasyu888, line 289: Just annotating - this is ORCA-205. Please self assign and pull into sprint.

src/orca/services/nextflowtower/ops.py
```diff
@@ -40,4 +46,87 @@ def workspace(self) -> str:
         if self.config.workspace is None:
             message = f"Config ({self.config}) does not specify a workspace."
             raise ConfigError(message)
-        return self.config.workspace
+        return self.config.workspace.lower()
+
+    def get_latest_compute_env(self, filter: Optional[str] = None) -> str:
```
thomasyu888, line 51: Just annotating - this is ORCA-166. Please self assign and pull into sprint.

src/orca/services/nextflowtower/models.py
```diff
@@ -0,0 +1,225 @@
...
+
+        Returns:
+            Workspace instance.
+        """
+        org = Organization.from_response(response)
+        return cls(response["workspaceId"], response["workspaceName"], org)
+
+
+@dataclass(kw_only=False)
+class LaunchInfo:
```
thomasyu888, line 91: Just a note, we may want to take a look at the [request body](https://tower.nf/openapi/index.html#post-/workflow/launch) and decide which ones we want to add in the future (you have the request body below as well). For example, "resume" might be a good one, but out of scope for this ticket.

src/orca/services/nextflowtower/models.py
```diff
@@ -0,0 +1,225 @@
...
+
+        Returns:
+            JSON representation.
+        """
+        output = {
+            "launch": {
+                "computeEnvId": self.compute_env_id,
+                "configProfiles": self.dedup(self.profiles),
+                "configText": self.nextflow_config,
+                "dateCreated": self.get_current_timestamp(),
```
thomasyu888: This is interesting, I didn't realize `dateCreated` was not a server-filled attribute.  Just curious - what are your thoughts on this sort of architecture - leaving it to the clients to fill out this information.

src/orca/services/nextflowtower/models.py
```diff
@@ -0,0 +1,225 @@
...
+
+    @staticmethod
+    def get_current_timestamp():
+        """Generate current timestamp using Tower's UTC format.
+
+        Returns:
+            Current timestamp.
+        """
+        return datetime.now().isoformat()[:-3] + "Z"
+
```
thomasyu888: Should there be a `from_response` dict here?  That said, the response object of a workflow launch is vastly different, so it might not make any sense.

src/orca/services/nextflowtower/client.py
```diff
@@ -131,56 +125,205 @@ def get(self, path: str, **kwargs) -> dict[str, Any]:
...
+        Args:
+            compute_env_id: Compute environment ID.
+            workspace_id: Tower workspace ID.
+
+        Returns:
+            Compute environment instance.
+        """
+        path = f"/compute-envs/{compute_env_id}"
+        params = self.generate_params(workspace_id, attributes="labels")
+        response = self.get(path, params=params)
```
thomasyu888: Just a note: I have to keep reminding myself that `response` is not actually a `requests.Response` object, but it's the json.

tests/services/nextflowtower/responses.py
```diff
@@ -41,25 +41,174 @@
...
+                {"IamRole": "foo"},
+                {"IamRole": "foo"},
+                {"IamInstanceProfile": "foo"},
+                {"Ec2LaunchTemplate": "TowerForge-5ykJF"},
+                {"BatchEnv": "foo"},
+                {"BatchQueue": "foo"},
+            ],
+            "discriminator": "aws-batch",
+        },
+        "dateCreated": "2023-04-26T00:49:49Z",
```
thomasyu888, line 157: This isn't isoformat which is causing a test to fail:
```
datetime.fromisoformat("2023-04-26T00:49:49Z")
# ValueError: Invalid isoformat string: '2023-04-26T00:49:49Z'
datetime.fromisoformat("2023-04-26T00:49:49")
# datetime.datetime(2023, 4, 26, 0, 49, 49)
```

tests/services/nextflowtower/test_ops.py
```diff
@@ -15,21 +18,121 @@ def test_for_an_error_when_the_workspace_attribute_is_missing(patch_os_environ):
...
+    # Create function for generating compute environments
+    def generate_compute_env(id, date):
+        compute_env_response = get_response("get_compute_env")["computeEnv"]
+        compute_env = models.ComputeEnv.from_response(compute_env_response)
+        return replace(compute_env, id=id, date_created=datetime.fromisoformat(date))
+
+    # Generate full compute environments
+    mocked_ops.client.get_compute_env.side_effect = [
+        generate_compute_env("5ykJF", "2023-04-26T00:49:49Z"),
+        generate_compute_env("9W2l7", "2023-01-11T13:32:54Z"),
```
thomasyu888, line 68: Same thing here with the isoformat strings
```suggestion
        generate_compute_env("5ykJF", "2023-04-26T00:49:49"),
        generate_compute_env("9W2l7", "2023-01-11T13:32:54"),
```

src/orca/services/nextflowtower/client.py
```diff
@@ -131,56 +125,205 @@ def get(self, path: str, **kwargs) -> dict[str, Any]:
...
+        Returns:
+            List of compute environments.
+        """
+        path = "/compute-envs"
+        params = self.generate_params(workspace_id, status=status)
+        response = self.get(path, params=params)
+        items = self.unwrap(response, "computeEnvs")
+        return [models.ComputeEnvSummary.from_response(item) for item in items]
+
+    def create_label(
```
thomasyu888, line 269: I could be wrong about this, but I'm not sure everyone actually has the ability to create labels.  It may be problematic if it is an entirely new workspace and someone tries to create the `launched-by-orca` label when they leverage this package to launch workflows.

src/orca/services/nextflowtower/ops.py
```diff
@@ -40,4 +46,87 @@ def workspace(self) -> str:
...
+        Args:
+            name: Label name.
+
+        Returns:
+            Label ID.
+        """
+        labels = self.client.list_labels(self.workspace_id)
+        for label in labels:
+            if label.resource:
+                continue
```
thomasyu888, line 96: I'm not fully understanding here, why is this a `continue`?  Even if we pass in a label that is a `resource`, don't we want to return the `label.id`?

This test: `test_that_create_label_ignores_resource_labels` is what spurs the discussion.  Should `create_label` be called if the resource label exists?

thomasyu888 (changes requested):
Awesome!  Thanks for putting this together.

src/orca/services/nextflowtower/client.py
BrunoGrandePhD: Done!

src/orca/services/nextflowtower/client.py
BrunoGrandePhD: I've been able to create labels with my personal user, which has the "launcher" role. This would be the minimum level of access needed by Orca if you want to launch workflows. If you try to launch a workflow with a read-only token, you should run into an unauthorized exception.

src/orca/services/nextflowtower/client.py
BrunoGrandePhD: I agree that it's a little confusing. Using `json` as the variable name should be clearer. I'll update.

src/orca/services/nextflowtower/client.py
BrunoGrandePhD: Not yet. I don't know if there are any paginated POST endpoints. As far as I know, pagination is only used for GET endpoints that return long lists of resources, whereas POST requests usually return info about the resource you just created.

src/orca/services/nextflowtower/client.py
BrunoGrandePhD: Fair enough. I'll simplify the code at the cost of introducing an extra HTTP request.

src/orca/services/nextflowtower/client.py
BrunoGrandePhD: It's a more consistent interface between paginated and non-paginated requests. If memory usage becomes an issue, we can revisit this.

src/orca/services/nextflowtower/models.py
BrunoGrandePhD: YAGNI. I try to avoid implementing methods until I need them. We might be able to use this model once we merge this branch with Brad's `get_workflow()` PR.

src/orca/services/nextflowtower/models.py
BrunoGrandePhD: I'm emulating the behavior that I'm seeing with the Tower web client. I'll try to launch a workflow without this field. If I can, then I'll drop this because it's simpler. IMHO, this feels like something that the backend should handle.

BrunoGrandePhD: I was able to run the `test_that_a_workflow_can_be_launched` integration test while setting this field to None, so I'm going to remove the timestamp code.

src/orca/services/nextflowtower/client.py
```diff
@@ -131,56 +125,205 @@ def get(self, path: str, **kwargs) -> dict[str, Any]:
...
+        See ``TowerClient.request`` for argument definitions.
 
-        Raises:
-            HTTPError: If the response lacks the expected keys.
+        Returns:
+            A dictionary from deserializing the JSON response.
+        """
+        return self.request_json("POST", path, **kwargs)
+
+    def unwrap(self, response: dict[str, Any], key: str) -> Any:
```
BWMac: Might it make sense to have functions like this one in a cross-service `utils` file? I can imagine this being useful for any service interacting with an API.

src/orca/services/nextflowtower/models.py
BrunoGrandePhD: You captured my intent with these models. Some of the models in the OpenAPI spec are very big. Replicating them here feels unnecessary, especially if we only use a small subset of the fields. As we need to add new features, we can expand what these models can store. 

I attempted a compromise with the `ComputeEnv` model by having a `raw` attribute that stores the full JSON response, so we'll see if that approach works well.

src/orca/services/nextflowtower/ops.py
BrunoGrandePhD: I'm trying to differentiate between workflow labels and resource tags, which Tower lumps together as labels. This function is intended to be about workflow labels. I updated the docstring to clarify. 

Note that a (non-resource) workflow label can coexist with a resource tag with the same name. Here, I was able to create a `CostCenter` label despite a `CostCenter` resource tag already existing:

```json
{
  "id": 89366116242820,
  "name": "CostCenter",
  "value": "000000",
  "resource": true
}

{
  "id": 266513533567473,
  "name": "CostCenter",
  "value": null,
  "resource": false
}
```

src/orca/services/nextflowtower/ops.py
BrunoGrandePhD: Done!

tests/services/nextflowtower/responses.py
BrunoGrandePhD: Interesting. This works fine on my machine:

```python
>>> from datetime import datetime
>>> datetime.fromisoformat("2023-04-26T00:49:49Z")
datetime.datetime(2023, 4, 26, 0, 49, 49, tzinfo=datetime.timezone.utc)
>>> datetime.fromisoformat("2023-04-26T00:49:49")
datetime.datetime(2023, 4, 26, 0, 49, 49)
```

I guess the issue is that this is technically RFC 3339 rather than ISO 8601. I implemented a simple parser utility function to handle these RFC 3339 timestamps from Tower. If this proves to be tricky going forward, we can look into using `pendulum` instead, which is already installed as a dependency of Airflow. I'm holding off for now as the current approach is more strict (I want to know if the format is consistent), and the `pendulum.parse()` function doesn't return a consistent type, so the code will be slightly more cluttered to avoid typing errors.

tests/services/nextflowtower/test_ops.py
BrunoGrandePhD: I updated the datetime parsing so that I can consistently use the same format adopted by Tower, which follows RFC 3339 (not ISO 8601).

src/orca/services/nextflowtower/client.py
BrunoGrandePhD: I agree. Once we add a second service where we need to implement a client (like Tower), we can abstract the common functionality to a base client class. Fortunately, most services that we do plan on using have clients (_e.g._ Synapse, SevenBridges). We might need to reuse this code once we implement support for GA4GH WES, but I don't think that's prioritized for the time being.

BrunoGrandePhD:
Thank goodness for the test suite! I was able to resolve multiple merge conflicts with confidence.

src/orca/services/nextflowtower/client.py
```diff
@@ -85,105 +86,241 @@ def request_json(self, method: str, path: str, **kwargs) -> dict[str, Any]:
...
+        self.update_kwarg(kwargs, "params", "max", 50)
+        self.update_kwarg(kwargs, "params", "offset", 0)
+
+        num_items = 0
+        all_items = list()
+        key_name = "items"  # Setting a default value
+        total_size = float("inf")  # Artificial value for initiating the while-loop
+        while num_items < total_size:
+            kwargs["params"]["offset"] = num_items
+            json = self.request_json(method, path, **kwargs)
```
thomasyu888, line 111: Nit: `json` as a variable conflicts with the base `json` package. It might be fine since we aren't importing it, but if we did decide to, this variable name might cause problems. Maybe `response_json` (I know it's long...)

src/orca/services/nextflowtower/ops.py
thomasyu888: Thanks for the clarification - I did not realize you could have resource and non-resource labels that are named the same.

thomasyu888 (approved):
🛸 Excellent!  Just a minor comment about the `json` variable, but going to pre-approve.

src/orca/services/nextflowtower/client.py
BrunoGrandePhD: Ah, that's probably why I didn't opt for `json` or `dict` at first. Let's stick with this for now. It's possible that we can avoid importing the `json` module where the `json` variable exists.

PR #16, opened 2023-04-28
[ORCA-164] adds `get_workflow` and `get_workflow_status` methods to ops

BWMac:
This PR adds the `get_workflow` and `get_workflow_status` methods to `NextflowTowerOps`. `get_workflow` takes the `workflow_id` for a run and returns all of the information that the API provides about it. `get_workflow_status` leverages `get_workflow`, but only returns a tuple: `(<workflow_status_string>, <workflow_complete_bool>)`. Tests for these methods are also included in `test_ops.py`.

I was struggling a bit with the mocking for `test_that_get_workflow_returns_expected_response`, but then I noticed that @BrunoGrandePhD had already solved my problem in #15, so I adopted his changes to `conftest.py`.

I tested the two new methods with the following code:
```
import os

from orca.services.nextflowtower.client_factory import NextflowTowerClientFactory
from orca.services.nextflowtower.config import NextflowTowerConfig
from orca.services.nextflowtower.ops import NextflowTowerOps

auth_token = os.environ.get("TOWER_ACCESS_TOKEN")
api_endpoint = "https://tower.sagebionetworks.org/api"
workspace = "Sage-Bionetworks/example-project"

print(ops.get_workflow(workflow_id="1Nozs3h3cT6Otb"))

print(ops.get_workflow_status(workflow_id="1Nozs3h3cT6Otb"))
```


src/orca/services/nextflowtower/ops.py
```diff
@@ -41,3 +41,32 @@ def workspace(self) -> str:
             message = f"Config ({self.config}) does not specify a workspace."
             raise ConfigError(message)
         return self.config.workspace
+
+    def get_workflow(self, workflow_id: str) -> dict:
```
thomasyu888: A point of discussion.  Should `get_workflow` be a separate function, or should we follow the [sevenbridges Ops class paradigm](https://github.com/Sage-Bionetworks-Workflows/py-orca/blob/56a40604f83c06d2cc07e503cc69033b832c8d2b/src/orca/services/sevenbridges/ops.py#L122)?

tests/services/nextflowtower/responses.py
```diff
@@ -63,3 +63,179 @@
...
+        "resume": True,
+        "success": True,
+        "logFile": None,
+        "outFile": None,
+        "operationId": None,
+        "ownerId": 28,
+    }
+}
+
+get_workflow_incomplete: dict = {
```
BrunoGrandePhD: If the main difference between this dictionary and the previous one is the `status` value, then I would only include the first one as `get_workflow` and change the status to `SUBMITTED` during the test to keep things DRY.

tests/services/nextflowtower/conftest.py
```diff
@@ -23,9 +23,14 @@ def config(patch_os_environ):
...
 @pytest.fixture
-def ops(config, client, mocker):
-    mock = mocker.patch.object(NextflowTowerOps, "client")
-    mock.return_value = client
+def ops(config):
+    yield NextflowTowerOps(config)
+
+
+@pytest.fixture
+def mocked_ops(config, client, mocker):
```
BrunoGrandePhD, line 31: Kudos for proactively checking out my PR for this solution!

src/orca/services/nextflowtower/ops.py
BrunoGrandePhD: My recommendation is that if a function is a simple wrapper around an API endpoint, then it should live in the `client` class. If the function provides additional interpretation over the API response or combines multiple API calls, then it should live in the `ops` class. 

Here, I think `get_workflow()` makes sense in the `client` class, whereas `get_workflow_status()` is better suited for the `ops` class since it provides higher-level functionality.

src/orca/services/nextflowtower/ops.py
```diff
@@ -41,3 +41,32 @@ def workspace(self) -> str:
...
+        Args:
+            workflow_id (str): The ID number for a workflow run to get information about
+
+        Returns:
+            tuple: Tuple containing 1. status (str) and
+            2. Whether the workflow is done (boolean)
+        """
+        response = self.get_workflow(workflow_id=workflow_id)
+        return (
+            response["workflow"]["status"],
```
BrunoGrandePhD: **Nit (_i.e._ [optional](https://google.github.io/eng-practices/review/reviewer/standard.html#:~:text=Reviewers%20should%20always%20feel%20free,they%20could%20choose%20to%20ignore.)):** I'm just wondering if we should create an enum for the Tower workflow status like we have for SevenBridges.

src/orca/services/nextflowtower/ops.py
```diff
@@ -41,3 +41,32 @@ def workspace(self) -> str:
...
+            bool(response["workflow"]["complete"]),
```
BrunoGrandePhD: **Nit (_i.e._ [optional](https://google.github.io/eng-practices/review/reviewer/standard.html#:~:text=Reviewers%20should%20always%20feel%20free,they%20could%20choose%20to%20ignore.)):** It might be slightly more readable to use `response["workflow"]["complete"] is not None`. But I don't really feel strongly on way or another. 

Did you confirm that failed workflows don't have this field filled in?

src/orca/services/nextflowtower/ops.py
```diff
@@ -41,3 +41,32 @@ def workspace(self) -> str:
...
+            workflow_id (str): The ID number for a workflow run to get information about
+
+        Returns:
+            response (dict): Dictionary containing information about the workflow run
+        """
+        path = f"/workflow/{workflow_id}"
+        response = self.client.get(path=path, params={"workspaceId": self.workspace_id})
+        return response
+
+    def get_workflow_status(self, workflow_id: str) -> tuple:
```
BrunoGrandePhD: Could you add a `TODO` here to consider switching to a `namedtuple` as the return value. This isn't urgent, so it can wait until the next time we touch this code. We would probably want to update the corresponding function in SevenBridges, which is why I'm suggesting that we do this later.

BrunoGrandePhD (changes requested):

src/orca/services/nextflowtower/ops.py
BWMac: Makes sense to me. I'll make that change.

src/orca/services/nextflowtower/ops.py
BWMac: I will make the first change, I agree it is a little more readable.

For failed workflows, it will still be filled in if the run has completed, but the returned tuple would read `("FAILED", True)`, so we would know that it failed and that the run is complete. Is there any more information that we need that this does not get at for this method?

src/orca/services/nextflowtower/ops.py
BWMac: I like this idea. Do you know where we can find an exhaustive list of all of the possible workflow statuses?

src/orca/services/nextflowtower/ops.py
BrunoGrandePhD: Here's the list of possible values for the workflow status:

```
SUBMITTED┃RUNNING┃SUCCEEDED┃FAILED┃CANCELLED┃UNKNOWN
```

I'm guessing that the first two prevent the `complete` field from being set, and the next three do result in the `complete` field being set. I'm not sure about `UNKNOWN` though. I hope that it does result in complete. 

I guess my hesitation is that we're relying on the implicit behavior of the `complete` field, whereas I think we would achieve better clarity if we just mapped each of the above statuses to whether or not it should be interpreted as done. 

As an example, SevenBridges has an enum for the workflow status, and it specifies which values are considered "terminal" ([source](https://github.com/sbg/sevenbridges-python/blob/2ad44789bd24b3e8803cb972c62bb612e22bce68/sevenbridges/models/enums.py#L61-L71)). This probably won't change the behavior, but it makes it more explicit and therefore more maintainable.

src/orca/services/nextflowtower/ops.py
BrunoGrandePhD: Yeah, it's in the Tower API ([source](https://tower-dev.sagebionetworks.org/api/openapi#get-/workflow)): 

```
SUBMITTED┃RUNNING┃SUCCEEDED┃FAILED┃CANCELLED┃UNKNOWN
```

image

src/orca/services/nextflowtower/ops.py
BWMac: OK, I can go ahead and try to add something like this. Should I assume `UNKNOWN` is a "terminal" status for now?

BrunoGrandePhD: Yeah, let's go with that for now.

BWMac: I implemented this in the latest commit, modeled off of the way the equivalent enum is used in the `sevenbridges` module. I also added tests for the new Nextflow Tower `TaskStatus` enum, not sure if it was needed or not.

BrunoGrandePhD (approved):
🚀

src/orca/services/nextflowtower/models/enums.py
```diff
@@ -0,0 +1,15 @@
+class TaskStatus:
```
thomasyu888: Nit: Should something like this be used? https://docs.python.org/3/library/enum.html

tests/services/nextflowtower/test_ops.py
```diff
@@ -33,3 +33,27 @@ def test_for_error_when_the_workspace_id_does_not_exist(ops, mocker, get_respons
...
+    mock.get_workflow.assert_called_once()
+    assert result == ("SUCCEEDED", True)
+
+
+def test_that_get_workflow_status_returns_expected_tuple_workflow_is_not_complete(
+    mocked_ops, mocker, get_response
+):
+    response = get_response("get_workflow")
+    response["workflow"]["complete"] = None
+    response["workflow"]["status"] = "SUBMITTED"
```
thomasyu888, line 54: Nit: I can imagine this using `pytest.parametrize` for all the supported statues and the expected output.

thomasyu888 (commented):
🔥  Awesome - Just some minor comments, I'll leave it to Bruno for final approval.  One question I had was should this wait for #15 ?

thomasyu888 (approved):

src/orca/services/nextflowtower/models/enums.py
BWMac: added in latest commit

tests/services/nextflowtower/test_ops.py
BWMac: Do you think that's necessary here? We are essentially just testing that values are grabbed from the `response` dictionary in this function. The other functionality that `pytest.parameterize` would test is if the `enum` is working correctly.

BWMac:
I can't merge anyway (don't have the necessary permissions) So I will let @BrunoGrandePhD merge this and #15 in whichever order.

tests/services/nextflowtower/test_ops.py
BrunoGrandePhD: We probably don't want to duplicate the logic for the various statuses. I think testing once for each type of status is good enough.

BrunoGrandePhD:
I'll merge this into `main` and update my branch in #15.

BrunoGrandePhD:
@BWMac: I just came across a failed workflow that doesn't have `complete` set. I'm glad we went with mapping the possible status values for determining workflow completion. 

```json
{
  "workflow": {
    "id": "3oW8mOkHmGlu4z",
    "ownerId": 28,
    "submit": "2023-05-03T18:27:00Z",
    "start": null,
    "complete": null,
    "dateCreated": "2023-05-03T18:27:00Z",
    "lastUpdated": "2023-05-03T18:34:26Z",
    "runName": "test_launch_workflow",
    "sessionId": "524d2bba-f4d3-4f38-8505-76db4928b527",
    "profile": "",
    "workDir": "s3://orca-service-test-project-tower-scratch/work",
    "commitId": null,
    "userName": "orca-service-account",
    "scriptId": null,
    "revision": null,
    "commandLine": "nextflow run nextflow-io/hello -name test_launch_workflow -params-file https://tower-dev.sagebionetworks.org/api/ephemeral/TwBprRYRKH6SYOsnQ61u8w.yaml -with-tower https://tower-dev.sagebionetworks.org/api",
    "projectName": "nextflow-io/hello",
    "scriptName": null,
    "launchId": "T7nyMWyt5WRLEoSWPRpQ",
    "status": "FAILED",
    "configFiles": null,
    "params": null,
    "configText": null,
    "manifest": null,
    "nextflow": null,
    "stats": null,
    "errorMessage": null,
    "errorReport": "Essential container in task exited\n\nDownloading plugin nf-amazon@1.11.3\nN E X T F L O W  ~  version 22.10.6\nPulling nextflow-io/hello ...\n downloaded from https://github.com/nextflow-io/hello.git\nCannot parse params file: https://tower-dev.sagebionetworks.org/api/ephemeral/TwBprRYRKH6SYOsnQ61u8w.yaml",
    "deleted": null,
    "peakLoadCpus": null,
    "peakLoadTasks": null,
    "peakLoadMemory": null,
    "projectDir": null,
    "homeDir": null,
    "container": null,
    "repository": null,
    "containerEngine": null,
    "scriptFile": null,
    "launchDir": null,
    "duration": null,
    "exitStatus": -1,
    "resume": false,
    "success": null
  },
  "progress": null,
  "orgId": 238103647815442,
  "orgName": "Sage-Bionetworks",
  "workspaceId": 177032410178845,
  "workspaceName": "orca-service-test-project",
  "labels": null,
  "starred": false,
  "optimized": null
}
```

BWMac:
@BrunoGrandePhD Interesting! I wonder what distinguishes a failed complete run from a failed incomplete run.

BrunoGrandePhD:
No idea!

PR #17, opened 2023-05-02
[WORKFLOWS-528] Allow `LaunchInfo` values to be missing

BrunoGrandePhD:
As I was working on launching workflows programmatically using py-orca for WORKFLOWS-528, I realized that the `LaunchInfo` model doesn't allow the `compute_env_id` to be missing. This prevents the user from creating a `LaunchInfo` instance without this information, with the intent of allowing the `ops.launch_workflow()` method to fill in these details.

### DIFF
--- a/src/orca/services/nextflowtower/models.py
+++ b/src/orca/services/nextflowtower/models.py
@@ -3,7 +3,7 @@ from collections.abc import Collection
 from dataclasses import field
 from datetime import datetime
 from enum import Enum
-from typing import Any, Optional
+from typing import Any, Iterable, Optional
 
 from pydantic.dataclasses import dataclass
 from typing_extensions import Self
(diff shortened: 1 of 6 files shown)


src/orca/services/nextflowtower/models.py
```diff
@@ -126,6 +126,34 @@ def fill_in(self, attr: str, value: Any):
...
+
+        Args:
+            attr: Attribute name.
+            values: New attribute values.
+        """
+        current_values = getattr(self, attr)
+        if not isinstance(current_values, list):
+            message = f"Attribute '{attr}' is not a list and cannot be extended."
+            raise ValueError(message)
+        updated_values = current_values + list(values)
```
thomasyu888, line 127: Does it make sense to run `dedup` here?

thomasyu888 (approved):
👍🏼 LGTM.  going to pre-approve, but just to confirm, does the `default_factory=list` make `profiles`, `user_secrets`, and etc optional fields as well?

BWMac (approved):

src/orca/services/nextflowtower/models.py
BrunoGrandePhD: Good idea!

BrunoGrandePhD:
> does the default_factory=list make profiles, user_secrets, and etc optional fields as well?

Yup, it initializes them with empty lists instead of None. I need to use a factory because lists are mutable. For more information, you can check out [this section](https://docs.python.org/3.11/library/dataclasses.html#mutable-default-values) in the Python docs on dataclasses.

PR #18, opened 2023-05-02
[ORCA-210] Add support for resuming workflows in NextflowTowerOps

BrunoGrandePhD:
Here's a summary of the changes introduced in this PR: 

- Add support for resuming workflows is to allow the `LaunchInfo` model to store the `resume` toggle and the `session_id`.
- Refactor the Nextflow Tower models to avoid redundancy with the `from_json()` method using a new `BaseTowerModel` base class
- Implement `list_workflows()` client and ops method
- Implement `get_previous_workflows()` and `get_latest_previous_workflow()` operations

I'm starting to be unhappy with the way mocking is currently being done. This is especially true for the mocked versions of the `launch_workflow()` operation in `test_ops.py`. I think I need to come up with fixtures that automatically patch the client methods to use the available JSON response in `responses.py`. I was going to start that here, but this PR is big enough as it is. I'll tackle this before I work on tests next time and dedicate its own PR.


src/orca/services/nextflowtower/client.py
```diff
@@ -83,7 +83,11 @@ def request_json(self, method: str, path: str, **kwargs) -> dict[str, Any]:
             A dictionary from deserializing the JSON response.
         """
         response = self.request(method, path, **kwargs)
-        response.raise_for_status()
+        try:
+            response.raise_for_status()
```
thomasyu888, line 87: Interesting... Is the nextflow API not returning the reasons in `response.reason`
https://requests.readthedocs.io/en/latest/_modules/requests/models/#Response.raise_for_status

src/orca/services/nextflowtower/models.py
```diff
@@ -233,20 +268,53 @@ class ComputeEnv(ComputeEnvSummary):
...
+class Workflow(BaseTowerModel):
+    """Nextflow Tower workflow run details."""
+
+    id: str
+    complete: Optional[datetime]
+    submit: Optional[datetime]
+    run_name: str
+    session_id: str
+    username: str
+    projectName: str
```
thomasyu888: Is camel case intentional here?

src/orca/services/nextflowtower/ops.py
```diff
@@ -104,17 +105,40 @@ def launch_workflow(
...
         Returns:
             Workflow run ID.
         """
+        # Make sure that essential attributes are set
+        if launch_info.pipeline is None or launch_info.run_name is None:
+            message = "LaunchInfo 'run_name' and 'pipeline' attributes must be set."
+            raise ValueError(message)
+
+        # Update launch_info if there are previous workflow runs
+        if not ignore_previous_runs:
```
thomasyu888, line 130: Ah, I just remembered for GENIE, we run a validation workflow on a daily basis which is the same pipeline, but the files downloaded are not always the same.  So couple of solutions:

* I think we would just specify: `ignore_previous_runs: True`
* Specify different run names

thoughts?

tests/services/nextflowtower/test_ops.py
thomasyu888, line 1: Do we need to test the `ignore_previous_runs` parameter?

src/orca/services/nextflowtower/utils.py
thomasyu888, line 1: Nice utility additions!

src/orca/services/nextflowtower/models.py
```diff
@@ -28,83 +30,153 @@ class TaskStatus(Enum):
...
-    id: int
-    username: str
-    email: str
+    Attributes:
+        _key_mapping: Mapping between Python and API key names.
+            Only discordant names need to be listed.
+    """
+
+    _: KW_ONLY
+    raw: Optional[dict[str, Any]] = field(default=None, repr=False, compare=False)
```
thomasyu888, line 42: Nice! Does this mean each class has a `.raw` attribute with the full json response?  

If so, can you speak to why you prefer this over creating kwargs? (e.g. `class.non_key_map_1`)

thomasyu888 (commented):
Nice work here!  An initial comment is that I have to remember to unset the `launch_label` class attribute if I don't have the `launch_by_orca` attribute set yet (at least testing some of these functions locally)

```
ops.launch_label = None
ops.list_workflows(search_filter="label:...")
```

tests/services/nextflowtower/test_ops.py
BrunoGrandePhD: I have an integration test that uses this parameter (`test_that_a_workflow_can_be_launched`).

src/orca/services/nextflowtower/models.py
BrunoGrandePhD: Nope, this was a copy-paste typo.

src/orca/services/nextflowtower/client.py
BrunoGrandePhD: The reason will be set based on the HTTP error code (_e.g._ 400 being "Bad Request"), but the response body will have additional information.

BrunoGrandePhD:
>  An initial comment is that I have to remember to unset the launch_label class attribute if I don't have the launch_by_orca attribute set yet (at least testing some of these functions locally)

If you're using the ops class, then yes. You can also use the client class to avoid this bit of functionality.

src/orca/services/nextflowtower/models.py
BrunoGrandePhD: > Does this mean each class has a .raw attribute with the full json response?

Yes, in case we need information from a response that isn't yet available through an attribute. Otherwise, you would have to issue a new py-orca release, which is annoying. 

> If so, can you speak to why you prefer this over creating kwargs? (e.g. class.non_key_map_1)

I'm not sure I understand the alternative you're proposing. Could you elaborate and/or give an example (maybe with some code)?

src/orca/services/nextflowtower/ops.py
BrunoGrandePhD: I would specify different run names. For example, you can name the runs based on the date (and time if run more than once per day). You could also set `ignore_previous_runs=True` to be more efficient (_i.e._ to avoid looking for previous workflow runs when you know there aren't any). However, that would prevent Orca from being able to re-run a failed workflow run. If the skipped files due to the failure would just get picked up the next day, then this isn't a concern, and you can set `ignore_previous_runs=True`.

src/orca/services/nextflowtower/models.py
thomasyu888: So instead of having things in `raw` just adding class attributes on the fly. Ie:

```
workflows = ops.list_workflows(search_filter="label:BPC")
workflow = workflows[1]
workflow.raw['commandLine'] # this exists today
workflow.commandLine # this doesn't
```

That said, I also don't think everything needs to exist as an attribute outside of "raw", just wanted to get your thoughts.

src/orca/services/nextflowtower/client.py
thomasyu888: Ah, makes sense. Thanks!

tests/services/nextflowtower/test_ops.py
thomasyu888: Thanks! I must have missed that!

thomasyu888 (approved):
🔥 I'm excited about this! Going to pre-approve - just one discussion point that I had.

src/orca/services/nextflowtower/models.py
BrunoGrandePhD: Ah, I see. Here's my rationale for my approach:

- Some (most?) responses are pretty long and often nested. For example, here's the [response](https://github.com/Sage-Bionetworks-Workflows/py-orca/blob/ccd3a5739be80c801d4737a6ab9bd39121055060/tests/services/nextflowtower/responses.py#L83-L169) when retrieving details on a compute environment. Exposing all of these values would be onerous, because each one has to be typed and mapped using `_key_mapping`. 
- I could override the `__getattr__()` method to fall back on the `raw` response if an instance attribute is missing, but I've had a hard time maintaining custom `__getattr__` logic in the past. Also, it would result in a mix of snake case and camel case. 
- We will generally only need a small subset of values in each HTTP response, so exposing all of them feels like premature optimization. YAGNI. 

If adding new attributes becomes burdensome, we can revisit this approach. I do appreciate you keeping me accountable and making sure that I've thought through my approach!

thomasyu888: Thanks for the thorough explanation - this makes sense to me!

BWMac (approved):

PR #19, opened 2023-05-04
[ORCA-217] Improve usability of Nextflow Tower service

BrunoGrandePhD:
As I was writing the [demonstration script](https://gist.github.com/BrunoGrandePhD/a12dddcedda331905e13f57bc83e1447), I found a few tweaks to improve the usability of `py-orca`: 

- Add some logging to be more transparent about what happens when you launch a workflow
- Construct configuration objects by default in `Ops` classes to conveniently pull from environment
- Document and test that all `Ops` class add a default factory for their `config` attributes
- Improve (slightly) the handling of the `WorkflowStatus` enum by inheriting from `str`
  - Starting in Python 3.11, you can use `StrEnum` for this, but it's too new to use in `py-orca`
- Fix bug with incrementing the suffix on the wrong run name
- Add a `get_workflow()` operation for the convenience of not having to provide the workspace ID

Here's an example run of the demo script: 

```console
$ pipenv run python3 demo.py
Loading .env environment variables...
Found an ongoing previous run: demo_8 (4F41pYrbgfXXGg)
Starting to monitor workflow: demo_8 (4F41pYrbgfXXGg)
Not done yet. Status is 'RUNNING'. Checking again in 5 min...
Not done yet. Status is 'RUNNING'. Checking again in 5 min...
Not done yet. Status is 'RUNNING'. Checking again in 5 min...
Not done yet. Status is 'RUNNING'. Checking again in 5 min...
Not done yet. Status is 'RUNNING'. Checking again in 5 min...
Done! Final status is 'SUCCEEDED'.
```

### DIFF
--- a/src/orca/services/nextflowtower/ops.py
+++ b/src/orca/services/nextflowtower/ops.py
@@ -1,8 +1,10 @@
+from dataclasses import field
 from functools import cached_property
 from typing import ClassVar, Optional
 
 from pydantic.dataclasses import dataclass
 
+from orca import logger
 from orca.errors import ConfigError
 from orca.services.base.ops import BaseOps
 from orca.services.nextflowtower.client import NextflowTowerClient
@@ -23,7 +25,7 @@ class NextflowTowerOps(BaseOps):
         client_factory_class: The class for constructing clients.
     """
 
-    config: NextflowTowerConfig
+    config: NextflowTowerConfig = field(default_factory=NextflowTowerConfig)
 
     client_factory_class = NextflowTowerClientFactory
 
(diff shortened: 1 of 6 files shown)


src/orca/services/nextflowtower/ops.py
```diff
@@ -1,8 +1,10 @@
+from dataclasses import field
 from functools import cached_property
 from typing import ClassVar, Optional
 
 from pydantic.dataclasses import dataclass
 
+from orca import logger
```
thomasyu888: Hmm.. @BWMac and I were discussing the logging module, and I've always read that the best practice is to instantiate a logger per module: https://stackoverflow.com/questions/15727420/using-logging-in-multiple-modules.

Thoughts?  

That said, I've also seen articles that say this: https://betterstack.com/community/questions/how-to-use-logging-in-multiple-modules/

src/orca/services/nextflowtower/ops.py
```diff
@@ -130,13 +132,20 @@ def launch_workflow(
         if not ignore_previous_runs:
             latest_run = self.get_latest_previous_workflow(launch_info)
             if latest_run:
+                status = latest_run.status.value
+                run_repr = f"{latest_run.run_name} ({latest_run.id})"
                 # Return ID for latest run if ongoing, succeeded, or cancelled
-                skip_statuses = {"SUCCEEDED", "CANCELLED"}
-                if not latest_run.is_done or latest_run.status.value in skip_statuses:
+                skip_statuses = {"SUCCEEDED"}
```
BWMac: Shouldn't this still include "CANCELLED"?

BWMac (changes requested):

src/orca/services/nextflowtower/ops.py
BrunoGrandePhD: I had included `CANCELLED` originally based on our conversation yesterday, but when I cancelled a workflow today for the demo, I had to delete that run to allow the script to work. So, maybe I'm not understanding why it should be included. How is it different from `FAILED`?

src/orca/services/nextflowtower/ops.py
BrunoGrandePhD: Mm, good catch! I'll work on updating this.

src/orca/services/nextflowtower/ops.py
BWMac: I guess I initially thought that a person's intention in canceling a workflow would be not to run that workflow at all, but it could definitely be that they realize they configured something incorrectly and then resuming would make sense. So on second thought, I agree with the decision to remove `CANCELLED` from this dictionary

BrunoGrandePhD: Thinking about this further, we might actually want to include `UNKNOWN` in this list because it might represent a workflow that's still running on AWS Batch, so we wouldn't want to relaunch it.

BWMac (approved):

PR #20, opened 2023-05-09
[ORCA-208] Extend all base classes to support Synapse service

BrunoGrandePhD:
The goal of this PR is to implement the service base classes for Synapse. This will make it easier to interact with Synapse in the context of Airflow thanks to the hook class.

### DIFF
--- a/src/orca/services/synapse/client_factory.py
+++ b/src/orca/services/synapse/client_factory.py
@@ -63,6 +63,8 @@ class SynapseClientFactory(BaseClientFactory):
             client: An authenticated client for this service.
         """
         profile = client.getUserProfile()
-        if profile["userName"] == "anonymous":
+        if profile.get("userName") == "anonymous":
             message = "Client is accessing Synapse anonymously."
             raise SynapseAuthenticationError(message)
+        else:  # pragma: no cover
+            pass  # To ignore lack of test coverage for happy path
--- a/src/orca/services/synapse/client_factory.py
+++ b/src/orca/services/synapse/client_factory.py
@@ -1,0 +1,68 @@
+from pydantic.dataclasses import dataclass
+from synapseclient import Synapse
+from synapseclient.core.exceptions import SynapseAuthenticationError
+
+from orca.errors import ConfigError
+from orca.services.base.client_factory import BaseClientFactory
+from orca.services.synapse.config import SynapseConfig
+
+
+@dataclass(kw_only=False)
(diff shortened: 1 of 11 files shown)


src/orca/services/synapse/client_factory.py
```diff
@@ -0,0 +1,70 @@
...
+        That is taken care of by the ``test_client()`` method.
+        That said, this method can raise an error if a response
+        is made but indicates a problem.
+
+        Args:
+            client: An authenticated client for this service.
+        """
+        profile = client.getUserProfile()
+        if profile.get("userName") == "anonymous":
+            message = "Client is accessing Synapse anonymously."
```
thomasyu888, line 67: Did we want to support anonymous for publicly accessible files? 

Although, as I type that out, that might only be 20% of the use cases so we can create a ticket to track.

src/orca/services/synapse/client_factory.py
```diff
@@ -0,0 +1,70 @@
...
+        auth_token = self.config.auth_token
+
+        if auth_token is None:
+            message = f"Config ({self.config}) is missing auth_token."
+            raise ConfigError(message)
+
+        # Ignore authentication error (leave that to `test_client_request`)
+        try:
+            client.login(authToken=auth_token)
+        except SynapseAuthenticationError:
```
thomasyu888, line 48: Can we check if there are other types of authentication errors thrown by the client?

I'm forgetting if all auth errors extend from this error class.

thomasyu888 (approved):
🔥 LGTM! I'm going to pre-approve, just some minor comments.

I haven't had a chance to test, but I trust that it works. I'll install and test when I get a chance.

src/orca/services/synapse/client_factory.py
BrunoGrandePhD: This error came up when I tried creating a client with a fake auth token (`foo`). If others come up here, it should be reported as a bug and then we can decide how to handle it. We might not want to ignore all exceptions here.

src/orca/services/synapse/client_factory.py
BrunoGrandePhD: Good point, but I agree with your follow-up point. We can create the ticket once someone asks for the feature. #YAGNI

BrunoGrandePhD:
I added a basic integration test to make sure that the interaction with Synapse actually works. I added a secret to this repo to enable integration tests in the CI/CD.

PR #21, opened 2023-05-16
[ORCA-228] Refactor workflow/task status as dataclasses

BrunoGrandePhD:
Before I add methods for monitoring workflows, I wanted to refactor the status tuple as a dataclass. This way, it's a little easier to pass around that information. I applied the change to both the Tower and SevenBridges services. 

For consistency, I'm using "state" to refer to the valid values like `"RUNNING"` and "status" for the convenience class, which includes an `is_done` property.


src/orca/services/sevenbridges/models.py
```diff
@@ -0,0 +1,23 @@
+from pydantic import validator
+from pydantic.dataclasses import dataclass
+from sevenbridges.models.enums import TaskStatus as TaskState
+
+
+@dataclass(kw_only=False)
+class TaskStatus:
```
thomasyu888, line 7: Probably not important now, but once we add more connectors, it could make sense to have a base status class.

thomasyu888 (approved):
🔥 LGTM!

src/orca/services/sevenbridges/models.py
BrunoGrandePhD: Agreed. We'll just have to think of how to handle the fact that the `state` attribute will consist of different enums.

PR #22, opened 2023-05-16
[ORCA-229] Implement `monitor_workflow()` for Tower

BrunoGrandePhD:
This PR migrates the [async function](https://github.com/Sage-Bionetworks-Workflows/ntap-add5-scripts/blob/8a24dea4402065be5fa65f523f9f6a2ea3160621/utils.py#L9-L22) that I was using in [ntap-add5-scripts](https://github.com/Sage-Bionetworks-Workflows/ntap-add5-scripts) to monitor workflow runs. 

While working on this, I realized that the refactoring that I did in #21 eliminated the need for `get_workflow_status()`, so I removed this function.

### DIFF
--- a/src/orca/services/nextflowtower/models.py
+++ b/src/orca/services/nextflowtower/models.py
@@ -321,6 +321,10 @@ class Workflow(BaseTowerModel):
         "state": "status",
     }
 
+    def __repr__(self) -> str:
+        """String representation of a workflow."""
+        return f"Workflow(run_name={self.run_name}, id={self.id}, state={self.state})"
+
     @property
     def status(self) -> WorkflowStatus:
         """Workflow run status."""
--- a/src/orca/services/nextflowtower/ops.py
+++ b/src/orca/services/nextflowtower/ops.py
@@ -1,3 +1,4 @@
+import asyncio
 import logging
 from dataclasses import field
 from functools import cached_property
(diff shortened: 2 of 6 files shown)


src/orca/services/nextflowtower/models.py
```diff
@@ -321,6 +321,10 @@ class Workflow(BaseTowerModel):
         "state": "status",
     }
 
+    def __repr__(self) -> str:
+        """String representation of a workflow."""
+        return f"Workflow(run_name={self.run_name}, id={self.id}, state={self.state})"
+
```
thomasyu888, line 327: Thoughts on adding a `__str__` representation too that is just the id?

src/orca/services/nextflowtower/ops.py
```diff
@@ -261,3 +250,25 @@ def get_latest_previous_workflow(
         # Otherwise, return latest based on submission timestamp
         sorted_runs = sorted(previous_runs, key=lambda x: x.get("submit"))
         return sorted_runs[-1]
+
```
thomasyu888, line 253: How will this work with the airflow sensor? Or do you envision airflow just to call:

workflow = ...get_workflow()
Status = workflow.status

thomasyu888 (approved):
🚀 LGTM!  going to pre-approve but just a couple of comments.

src/orca/services/nextflowtower/models.py
BrunoGrandePhD: Dataclasses already have default implementations for `__repr()__` and `__str__()`. I'm just overriding this one because the full output is too long for logging, which I want both the human-friendly run name and computer-friendly ID to appear. I'm aware of the `field(..., repr=False)` parameter, but I don't want to set that for all but a handful of attributes.

src/orca/services/nextflowtower/ops.py
BrunoGrandePhD: It would look like this ([original version](https://github.com/Sage-Bionetworks-Workflows/orca-recipes-airflow/blob/0ef9ccbd958e691f37f2d99bc19b10c7661fa00a/dags/cavatica-launch-poc-v2.py#LL49C1-L61C53)):

```python
@task.sensor(poke_interval=10, timeout=604800, mode="poke")
def monitor_workflow(params, workflow_id):
    hook = NextflowTowerHook(context["params"]["conn_id"])
    workflow = hook.ops.get_workflow(workflow_id)
    return PokeReturnValue(workflow.is_done, workflow.state)
```

Or the class-based equivalent of the above.

PR #24, opened 2023-05-30
[ORCA-209] Monitor Submission View

BWMac:
This PR adds the method `monitor_submission_view` to `SynapseOps`.

The function simply queries a submission view id to retrieve rows where the submission `status` is `RECEIVED`, reads the rows, and returns `TRUE` if the number of rows is greater than 0 or `FALSE` if not.

This will be used in the initial step of a submission evaluation DAG, where the DAG will only execute fully if this step returns `TRUE`.

### DIFF
--- a/src/orca/services/synapse/ops.py
+++ b/src/orca/services/synapse/ops.py
@@ -38,6 +38,8 @@ class SynapseOps(BaseOps):
         Returns:
             True if there are "RECEIVED" submissions, False otherwise.
         """
-        received_submissions = syn.getSubmissionBundles(evaluation_id, status="RECEIVED")
+        received_submissions = self.client.getSubmissionBundles(
+            evaluation_id, status="RECEIVED"
+        )
         submissions_num = sum(1 for submission in received_submissions)
         return submissions_num > 0
--- a/src/orca/services/synapse/ops.py
+++ b/src/orca/services/synapse/ops.py
@@ -31,7 +31,7 @@ class SynapseOps(BaseOps):
     client_factory_class = SynapseClientFactory
 
     client: ClassVar[Synapse]
-      
+
     @cached_property
     def fs(self) -> SynapseFS:
         """Synapse file system."""
@@ -43,7 +43,6 @@ class SynapseOps(BaseOps):
 
         return SynapseFS(auth_token=auth_token)
 
-   
     def monitor_evaluation_queue(self, evaluation_id: str) -> bool:
         """Monitor an evaluation queue in Synapse.
 
@@ -58,4 +57,3 @@ class SynapseOps(BaseOps):
         )
         submissions_num = sum(1 for submission in received_submissions)
         return submissions_num > 0
-
--- a/src/orca/services/synapse/ops.py
+++ b/src/orca/services/synapse/ops.py
@@ -29,17 +29,15 @@ class SynapseOps(BaseOps):
 
     client: ClassVar[Synapse]
 
-    def monitor_submission_view(self, view_id: str) -> bool:
-        """Monitor a submission view for completion.
+    def monitor_evaluation_queue(self, evaluation_id: str) -> bool:
+        """Monitor an evaluation queue in Synapse.
 
         Args:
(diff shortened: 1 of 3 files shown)


src/orca/services/synapse/ops.py
```diff
@@ -28,3 +28,18 @@ class SynapseOps(BaseOps):
...
+    def monitor_submission_view(self, view_id: str) -> bool:
+        """Monitor a submission view for completion.
+
+        Args:
+            view_id: The ID of the view to monitor.
+
+        Returns:
+            True if there are new "RECEIVED" submissions, False otherwise.
+        """
+        received_submissions = self.client.tableQuery(
```
thomasyu888: The most annoying part about views in Synapse in general, is that when you query them, it triggers an indexing job.  So here's what happens (at least when I was working with them)

1. Submission view is empty
2. Someone submits
3. You query the submission view -> it still shows up as empty, BUT the backend triggers an indexing of the view.
4. you query the view again -> the new submission shows up

BWMac: So I should have it query twice every time? I have noticed a quirk where I have to refresh the page twice to see an updated status in the view in the UI...

Or maybe we just have it check very frequently (I think you said every couple of minutes previously) and it will still grab the submission pretty soon after they are pushed?

thomasyu888: The quirk you are seeing is what I had outlined - Status updates are treated the same.  Some solutions (some you've stated already)

1. Do two queries:

    ```
    "select * from synNNNN limit 1"
    ```

1. Run it very frequently (every 5 minutes is probably fine - just annoying to test)
1. Use `syn.getSubmissionBundle(evaluationid, status='RECEIVED')` which always grabs new submissions.

I don't feel strongly about which one we pick, but the 2nd one introduces the "longest lag time" depending on how frequently these DAGs are run.

BWMac: I think option 3 makes the most sense, but I am getting this error when I try to use it:
```
SynapseHTTPError: 403 Client Error: 
User lacks READ_PRIVATE_SUBMISSION access to Evaluation syn51356905
```
Even though I have administrator access to the submissions and the project as a whole

BWMac: I updated the function to use `syn.getSubmissionBundles`. Querying twice feels silly and relying on the DAG running frequently is not ideal.

I will add a test for this function before marking this "ready for review".

thomasyu888: Agree with your rationale here - nice!

src/orca/services/synapse/ops.py
```diff
@@ -42,3 +42,18 @@ def fs(self) -> SynapseFS:
...
+        Args:
+            evaluation_id: The Synapse ID of the queue to monitor.
+
+        Returns:
+            True if there are "RECEIVED" submissions, False otherwise.
+        """
+        received_submissions = self.client.getSubmissionBundles(
+            evaluation_id, status="RECEIVED"
+        )
+        submissions_num = sum(1 for submission in received_submissions)
```
thomasyu888, line 58: Nit: So... something that might be more efficient is doing: 

```
next(received_submissions)
```

So that you aren't looping through everything

BWMac: `next()` on an empty iterator (no new submissions) causes an error though, so I would need to use try/except like this:
```
received_submissions = self.client.getSubmissionBundles(
    evaluation_id, status="RECEIVED"
)
try:
     next(received_submissions)
     return True
except StopIteration:
     return False
```

thomasyu888: Ah, I see. lets move forward with what you have since there shouldn't be too many submissions in a RECEIVED state at any time anyhow.

thomasyu888 (approved):
🪐 Awesome!  LGTM!

BWMac:
@thomasyu888 Thanks for approving! Just a heads up that I can't merge this myself. And I will need this functionality to complete my next task.

PR #25, opened 2023-06-15
Improve `.env.example` template

thomasyu888:
We have to `source` the `.env` file and include `export`


BWMac (approved):
LGTM. I re-read the instructions for the demo script in the README and I think these changes along with the existing instructions are enough for people to get the hang of it.

PR #26, opened 2023-10-12
[ORCA-280] Adds Support for Nextflow Entrypoints in `LaunchInfo`

BWMac:
This PR adds support for Nextflow entrypoints to be passed via `LaunchInfo` when `launch_workflow` is run. This is done simply by adding the `entry_name` variable to the `LaunchInfo` class. `entry_name` is optional when `LaunchInfo` is instantiated and defaults to `""` rather than `None` to avoid needing to hard-code an update during the `launch_workflow` call. This change has been tested using variations on the following script (and with environment variables configured similar to those needed to run the `demo.py` script in this repo):
```
from orca.services.nextflowtower import NextflowTowerOps
from orca.services.nextflowtower.models import LaunchInfo
import os

ops = NextflowTowerOps()

info = LaunchInfo(
    run_name="model_to_data_orca_test_entry_set",
    pipeline="Sage-Bionetworks-Workflows/nf-synapse-challenge",
    profiles=["tower"],
    revision="main",
    entry_name="MODEL_TO_DATA_CHALLENGE",
    workspace_secrets=["SYNAPSE_AUTH_TOKEN"],
)

ops.launch_workflow(info, "ondemand")
```
Nextflow Tower test runs:
- [Workflow that supports entrypoints with `entry_name` set](https://tower.sagebionetworks.org/orgs/Sage-Bionetworks/workspaces/example-project/watch/3eXqRdwfBdRAeu)
- [Workflow that supports entrypoints with `entry_name` not set](https://tower.sagebionetworks.org/orgs/Sage-Bionetworks/workspaces/example-project/watch/qdgPiCFGP7HGQ)
__Note:__ When `entry_name` is not set and there is no default workflow in `main.nf`, the run will finish "successfully", but nothing actually happened. We may want to include something that catches when `entry_name` is not set in `nf-synapse-challenge`.
- [Workflow that does not support entrypoints with `entry_name` not set](https://tower.sagebionetworks.org/orgs/Sage-Bionetworks/workspaces/example-project/watch/1ljP4jfJ8YQ1D0)

### DIFF
--- a/src/orca/services/nextflowtower/models.py
+++ b/src/orca/services/nextflowtower/models.py
@@ -167,6 +167,7 @@ class LaunchInfo(BaseTowerModel):
     label_ids: list[int] = field(default_factory=list)
     resume: bool = False
     session_id: Optional[str] = None
+    entry_name: Optional[str] = ""
 
     @root_validator()
     def check_resume_and_session_id(cls, values: dict[str, Any]):
@@ -223,7 +224,7 @@ class LaunchInfo(BaseTowerModel):
             "configProfiles": dedup(self.profiles),
             "configText": self.nextflow_config,
             "dateCreated": None,
-            "entryName": None,
+            "entryName": self.get("entry_name"),
             "headJobCpus": None,
             "headJobMemoryMb": None,
             "id": None,


BryanFauble (approved):
Changes LGTM!

fyi: When I was trying to look in nextflow tower I get:
image
Could I not have all the required permissions? Do you need to share the project out?

BWMac:
@BryanFauble You must not have access to the `example-project` Tower workspace yet.

@thomasyu888 We should probably add access for Bryan.

src/orca/services/nextflowtower/models.py
```diff
@@ -223,7 +224,7 @@ def to_json(self) -> dict[str, Any]:
             "configProfiles": dedup(self.profiles),
             "configText": self.nextflow_config,
             "dateCreated": None,
-            "entryName": None,
```
thomasyu888, line 226: I see that this used to be None, can you clarify the hardcoding bit when setting this value to be "".

BWMac: If I set `entry_name: Optional[str] = None` instead of `""`, if you aren't using the `entry_name` variable when you create your `LaunchInfo` object it triggers [this](https://github.com/Sage-Bionetworks-Workflows/py-orca/blob/9a8483a5a30ac2c5174ea228b26eeb2dcf23607f/src/orca/services/nextflowtower/models.py#L96) error when you try to launch the workflow, unless you include a line similar to 
```
launch_info.fill_in("compute_env_id", compute_env.id)
```
for `entry_name` in `launch_workflow`. But since that line would look like
```
launch_info.fill_in("entry_name", "")
```
I figured it was better to make the default `""`.

thomasyu888: Ahh I see. Makes sense! Thanks for the explanation.

thomasyu888 (approved):
🔥 LGTM

PR #27, opened 2023-10-16
[ORCA-282] Adds "Keep Alive" CI Step

BWMac:
This PR introduces a CI step to prevent our scheduled workflow job from being paused due to the repository not being updated for >=60days. This step uses the [Keep Alive Action](https://github.com/marketplace/actions/keepalive-workflow). This Action will run each time our scheduled job does, and if it detects that this repo has not been updated for >=44 days it will push a dummy commit thus preventing our job from being paused after the 60-day limit is hit.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,8 +55,8 @@ jobs:
         # Prevent scheduled workflow from ever being paused
       - name: Keepalive Workflow
         uses: gautamkrishnar/keepalive-workflow@1.1.0
-        
-                    
+
+
   test:
     needs: prepare
     strategy:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -52,7 +52,11 @@ jobs:
           name: python-distribution-files
           path: dist/
           retention-days: 1
-
+        # Prevent scheduled workflow from ever being paused
```
thomasyu888, line 55: Interesting. Do we know what this action does to keep the workflow alive?

BWMac: It creates a dummy commit if the repo is >50 days stale when the scheduled job runs

BWMac: I'm updating it to do the dummy commit at 44 days instead. I realized that if 45-49 days had passed when a scheduled job ran it would not do the dummy commit before the 60-day limit.

thomasyu888 (approved):
🔥 LGTM!

PR #29, opened 2023-10-24
[IBCDPE-691] Adds Ability to Retrieve Workflow Task Logs from Tower

BWMac:
This PR adds functionality to the `nextflowtower` service to retrieve metadata for the individual tasks in a Tower workflow run, and then to retrieve the execution logs for those individual tasks. 

To do this, I implemented the `get_workflow_tasks` methods in `NextflowTowerClient` and `NextflowTowerOps`. `client.get_workflow_tasks` uses the `/workflow/{workflow_id}/tasks` Tower API endpoint to retrieve information about all of the tasks. Each individual task dictionary is then converted into a `models.WorkflowTask` object. 

Task logs can now be retrieved by using the `get_task_logs` methods in `client` and `ops`. `client.get_task_logs` uses the `/workflow/{workflow_id}/log/{task_id}` Tower API endpoint to retrieve execution logs for an individual task. 

Tests and example responses have also been added where necessary for all new methods.

Finally, I have included `demo_nf_logs.py` as an example of a Metaflow DAG that users can follow to retrieve task logs and evaluate why a workflow task has failed. I am unsure whether this script should belong in this repo, or in `orca-recipes`, however.


src/orca/services/nextflowtower/client.py
```diff
@@ -113,7 +113,9 @@ def request_paged(self, method: str, path: str, **kwargs) -> dict[str, Any]:
         while num_items < total_size:
             kwargs["params"]["offset"] = num_items
             json = self.request_json(method, path, **kwargs)
-            total_size = json.pop("totalSize")
+            total_size = (
```
BWMac: This change and the one just below were made because the `tasks` API endpoint response JSON uses `total` while the paginating endpoints we were already using use `totalSize`. `request_paged` was not being called for `get_workflow_tasks` before.

src/orca/services/nextflowtower/client.py
```diff
@@ -369,3 +371,47 @@ def list_workflows(
...
+        Returns:
+            List of WorkflowTask objects.
+        """
+        path = f"/workflow/{workflow_id}/tasks"
+        params = self.generate_params(workspace_id)
+        json = self.get(path=path, params=params)
+        items = self.unwrap(json, "tasks")
+        return [models.WorkflowTask.from_json(item["task"]) for item in items]
+
+    def get_task_logs(
```
BWMac, line 395: One weakness of this feature is that you have to make 1 API call per task to get the logs. This could make it difficult to retrieve thousands of logs in a timely fashion without parallelization. For example, at present it takes ~10 minutes for `demo_nf_logs.py` to run on a workflow with 273 tasks.

demo_nf_logs.py
```diff
@@ -0,0 +1,86 @@
+import asyncio
```
BWMac: I'm wondering if this example script would make more sense as a local recipe in `orca-recipes`

demo_nf_logs.py
```diff
@@ -0,0 +1,94 @@
...
+    )
+
+    # Example error reasons for nf-histoqc
+    log_reasons = [
+        "NO tissue remains detectable!",
+        "NO tissue remains detectable in mask!",
+        "Inappropriate argument value (of correct type)",
+    ]
+
+    async def get_logs(self, task):
```
BryanFauble: Nit: add type hints

src/orca/services/nextflowtower/client.py
BryanFauble: if `json.get("totalSize")` is 0 it'll do a `json.pop("total")`. I'm not sure if that is intended in this case.

BWMac: The response `json` will have either `"totalSize"` or `"total"`, not both. Those are the cases I was trying to allow for.

BryanFauble: If they have either or - this would throw a KeyError if "totalSize" = 0 since "total" is not present

src/orca/services/nextflowtower/client.py
```diff
@@ -134,7 +136,7 @@ def get(self, path: str, **kwargs) -> dict[str, Any]:
             A dictionary from deserializing the JSON response.
         """
         json = self.request_json("GET", path, **kwargs)
```
BryanFauble, line 136: I think there is room for improvement here (And correct me if I am wrong). The way I understand this is currently working:

1. `request_json` is called to get the first set (Or all results if the count is low enough)
2. If there are enough results you need to drop to your paginated result set
3. You get the first set again (You already got it in NUM: 1), and then use the offset param to get all subsequent results.

Ideally you should retain the items you originally got and start at that offset in the `request_paged` call

src/orca/services/nextflowtower/client.py
BWMac: addressed in latest commit

BryanFauble: I think chaining .get would work:
```
json = {
    "totalSize": 1,
    "total": 2
}

# Print 1
print(json.get("totalSize", json.get("total", 0)))

json = {
    "total": 2
}

# Print 2
print(json.get("totalSize", json.get("total", 0)))

json = {}

# Print 0
print(json.get("totalSize", json.get("total", 0)))
```

BWMac: Chaining .get actually breaks a whole bunch of tests, I guess either it doesn't like the default 0 or the `pop` removing the dictionary key is needed

src/orca/services/nextflowtower/client.py
BWMac: @BryanFauble and I had a quick huddle to discuss this, and then after toying around with some small changes I think if we are going to implement a change to make this more efficient it will need to be a separate ticket due to its potential to impact so much of the `nextflowtower` service (out of scope for this ticket)

src/orca/services/nextflowtower/client.py
BWMac: What about 
[`json.pop("totalSize", None) or json.pop("total", 0)`](https://github.com/Sage-Bionetworks-Workflows/py-orca/blob/cdfdad2fb74db14743bc230f78da35ba3e81c239/src/orca/services/nextflowtower/client.py#L116)

BryanFauble (approved):
Changes LGTM!

demo_nf_logs.py
thomasyu888: Yes, let's add it to orca recipes!

src/orca/services/nextflowtower/client.py
thomasyu888: Would the other method be to just get the logs directly via AWS API?

tests/services/nextflowtower/responses.py
```diff
@@ -527,3 +527,249 @@
     ],
     "totalSize": 3,
 }
+
```
thomasyu888, line 530: Out of curiosity, did you try and leverage copilot to generate this test response?

thomasyu888 (approved):
🔥

src/orca/services/nextflowtower/client.py
BWMac: Yes, I did look into this alternative method:

- get tasks using `get_workflow_tasks`
- loop though tasks to get `work_dir` and use `s3fs` or another AWS-using method to download `command.log` from the bucket
- read contents of files into memory to process strings

But, I found this to be at least as inefficient if not worse. Mainly because of the need to actually download files which would require a more complicated implementation since we would not want the potential for thousands of files to be downloaded to a local disk at one time. In addition to needing to handle the case where the file is not there at all which does happen in failed Tower workflow processes depending on how/when they fail.

tests/services/nextflowtower/responses.py
BWMac: I did not, I used an existing Hello World workflow run to get the example response. I will definitely consider doing that in the future!

demo_nf_logs.py
BWMac: I have removed the script and `pandas` from the `dev` dependencies of this repo and opened [this](https://github.com/Sage-Bionetworks-Workflows/orca-recipes/pull/28) PR in `orca-recipes` to add the script and bump the `py-orca` version. The `orca-recipes` PR will be finalized after this one has been merged and a new version has been minted.

PR #30, opened 2023-11-07
Correct pipfile and add readme note

BryanFauble:
Problem:
In the pipfile the name of the package does not match the name defined in the setup.cfg. It leads to the error below:
```
Installing dependencies from Pipfile.lock (e4bbd8)...
[pipenv.exceptions.InstallError]: ERROR: Could not find a version that satisfies the requirement py-orca[all]==1.3.2.post1.dev19+gcdfdad2.d20231026 (from versions: 1.0.0, 1.0.1, 1.1.0, 1.2.0, 1.3.0, 1.3.1, 1.3.2, 1.3.3)
[pipenv.exceptions.InstallError]: ERROR: No matching distribution found for py-orca[all]==1.3.2.post1.dev19+gcdfdad2.d20231026
ERROR: Couldn't install package: {}
 Package installation failed...
```

Solution:
Correct the name of the package in the pipfile

Testing:
I verified I could use `pipfile install -d` and all dependencies were installed as expected

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -4,10 +4,10 @@ verify_ssl = true
 name = "pypi"
 
 [packages]
-orca = {editable = true, path = ".", extras = ["all"]}
+py-orca = {editable = true, path = ".", extras = ["all"]}
 
 [dev-packages]
-orca = {editable = true, path = ".", extras = ["testing", "dev"]}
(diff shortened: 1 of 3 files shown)


BWMac (approved):
Interesting. This was always working before so I wonder if there is something about the newer versions of `pipenv` that causes this problem.

PR #31, opened 2023-12-13
[ORCA-294] Fix `tox` implementation in CI pipeline

BWMac:
**Problem:**

Our CI pipeline is failing to test our code with the intended Python versions.

**Solution:**

Update ci.yml to use the tox-gh GH Actions plugin.

Notes:

- After a discussion with @thomasyu888, we determined that for now, we will only support Python 3.10 and 3.11 in `py-orca`
     - The dependency issues with this repository and Python 3.8/3.9 have been very difficult to navigate, primarily because of `apache-airflow`.
     - Once [ORCA-293](https://sagebionetworks.jira.com/browse/ORCA-293) is complete, we will try to support Python 3.8 and 3.9 again if possible.
     - Although this is not ideal, it should not impact the use of `py-orca` much, as it is primarily used within `orca-recipes`, which uses Python 3.10 by default.
- A GH Action run with these changes can be viewed [here](https://github.com/Sage-Bionetworks-Workflows/py-orca/actions/runs/7214575676/job/19657012933).
     - Please verify that the versions listed in the logs for the Run tests steps under tox:py match those expected before approving as a sanity check.

[ORCA-293]: https://sagebionetworks.jira.com/browse/ORCA-293?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,8 +36,10 @@ jobs:
         with: {python-version: "3.11"}
       - name: Run static analysis and format checkers
         run: pipx run pre-commit run --all-files --show-diff-on-failure
+      - name: Install tox-gh plugin
+        run: python -m pip install tox-gh>=1.2
       - name: Build package distribution files
-        run: pipx run --spec 'tox~=3.0' tox -e clean,build
+        run: tox -e clean,build
       - name: Record the paths of wheel and source tarball distributions
         id: distribution-paths
         run: |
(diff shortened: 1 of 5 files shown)


thomasyu888 (approved):
🔥 LGTM! As discussed, there's another ticket to tackle the constraints issue

PR #32, opened 2024-01-03
CI testing

BWMac:
**Problem:**

Recently, our Airflow deployment has begun to fail when some un-constrained dependencies are updated. Our testing pipeline in `py-orca` has also begun to fail due to the same issue.

**Solution:**

Update `py-orca` to use the constraints [file](https://raw.githubusercontent.com/apache/airflow/constraints-2.7.2/constraints-3.10.txt) provided by the airflow developers when installing dependencies.

Notes:

- Because dependency constraints files are currently only supported in `pip`, I needed to remove our previous `pipenv` setup.
- Development environment setup is now achieved by running the `dev_setup.sh` script which creates a Python virtual environment, upgrades `pip`,  installs airflow with its constraints, and installs `py-orca` with all of its extras.
- To use these constraints in our CI pipeline, I added the `requirements-airflow.txt` file as a dependency for the `testenv` section in `tox.ini`. This ensures that the required versions of airflow dependencies are installed in the Github Action runner.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -109,6 +109,8 @@ This often provides additional considerations and avoids unnecessary work.
    source py-orca-venv-<your_python_version>/bin/activate
    ```
 
+   **Note:** `py-orca` currently supports Python versions >=3.10 and <3.12.
+
 5. Install [pre-commit] hooks:
 
    ```
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -101,19 +101,18 @@ This often provides additional considerations and avoids unnecessary work.
    cd orca
    ```
 
-4. Install `pipx` to easily run Python CLI tools like `tox` and `pipenv`.
-
 4. Create an isolated virtual environment containing package dependencies,
    including those needed for development (*e.g.* testing, documentation) by running:
 
    ```console
-   pipx run tox -e pipenv
+   ./dev_setup.sh <your_python_version>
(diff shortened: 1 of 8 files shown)


CONTRIBUTING.md
```diff
@@ -101,19 +101,18 @@ This often provides additional considerations and avoids unnecessary work.
...
    ```
 
-4. Install `pipx` to easily run Python CLI tools like `tox` and `pipenv`.
-
 4. Create an isolated virtual environment containing package dependencies,
    including those needed for development (*e.g.* testing, documentation) by running:
 
    ```console
-   pipx run tox -e pipenv
+   ./dev_setup.sh <your_python_version>
```
BryanFauble: Is there a recommended python version to use here for development? Maybe "use the lowest version this project officially supports"?

BWMac: I was planning on just leaving it as use any supported version, which is enforced in the `dev_setup.sh` script. Is there a strong reason to make a specific recommendation?

Also, would you mind holding off on any further review until I mark this as ready for review?

BryanFauble: > I was planning on just leaving it as use any supported version, which is enforced in the dev_setup.sh script.

Works for me

BryanFauble (approved):
Changes look go to me. I was able to follow the contributing docs, run the shell script, and run pytest on the unit tests in a new environment.

thomasyu888 (approved):
💯 Nice!

PR #33, opened 2024-01-03
Add code owners and update authors

thomasyu888:
Add CODEOWNERS for auto code reviewers and update author list


BWMac (approved):
LGTM. I don't think tests are going to pass in CI/CD because of the dependency issue though.

PR #35, opened 2024-01-12
[ORCA-297] Fixing tutorial demo (`demo.py` and documentation)

jaymedina:
#### problem 

The initial problem was the `demo.py` was failing at step: `monitor_synstage` due to it still pointing at the old `nf-synstage` workflow which runs with DSL1 which is no longer supported:

image

Further blockers involved failures at step `monitor_rnaseq`. More context in **solution** below.

----
#### solution

- [x] Point `demo.py` to the new `synstage.nf` workflow in the `nf-synapse` repository.
- [x] Fix S3 AccessDenied error by changing `s3_prefix` in example to a bucket that DPE tower/tower-dev workspaces have access to, and adding a caveat that the s3_prefix should be accessible by the tower workspace you're launching from.
- [x] Mention in documentation that `NF_SYNSTAGE` workflow runs should have unique names (i.e. no consecutive runs of `my_test_dataset_synstage` with a `_#` tacked on at the end). Otherwise, this will lead to consecutive steps breaking because the samplesheet can't be found.
- [x] Co-authored with @BWMac : Fix incorrect directory path in `get_staged_samplesheet`
- [x] Update documentation in README.md and demo.py

----
#### testing & preview

With the changes noted above, all steps up until `monitor_synindex` pass, which is expected behavior ✅ : 

image

[source](https://tower-dev.sagebionetworks.org/orgs/Sage-Bionetworks/workspaces/example-dev-project/watch/59Y4Ffcv3CjgZ0)

----
#### out of scope

It has been discovered that the current `demo.py` is not written to support getting staged samplesheets from consecutive runs of `synstage` with the same run name, due to the way the names are updated (e.g. `my_test_dataset_synstage_2`, `my_test_dataset_synstage_3`, etc.

This is out of scope but is a nice-to-have, low priority item that can be moved to the bottom of the backlog. In the meantime, documentation will be updated in this PR to notify users that runs of `NF_SYNSTAGE` should have unique names.

----
This is a blocker for [IBCDPE-749](https://sagebionetworks.jira.com/browse/IBCDPE-749) and therefore high in urgency 🔥 

[IBCDPE-749]: https://sagebionetworks.jira.com/browse/IBCDPE-749?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ This Python package provides the components to connect various third-party servi
 
 ## Demonstration Script
 
-This repository includes a demonstration script called [`demo.py`](demo.py), which showcases how you can use `py-orca` to launch and monitor your workflows on Nextflow Tower. Specifically, it illustrates how to process an RNA-seq dataset using a series of workflow runs, namely `nf-synstage`, `nf-core/rnaseq`, and `nf-synindex`. `py-orca` can be used with any Python-compatible workflow management system to orchestrate each step (_e.g._ Airflow, Prefect, Dagster). The demonstration script uses [Metaflow](https://metaflow.org/) because it's easy to run locally and has an intuitive syntax.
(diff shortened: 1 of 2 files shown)


thomasyu888 (approved):
🔥 Nice!

thomasyu888 (approved):
🔥 this is great!  If my memory serves me right, the synindex steps fails right now - eventually we will make it all work!

BWMac (approved):

PR #37, opened 2024-02-27
[ORCA-306] Get and update Challenge submissions

jaymedina:
### problem

Helper methods needed to be added for basic operations with Synapse, such as retrieving the submissions with a certain submission status and updating submission statuses. This is for the new Sage Bionetworks Challenges backend workflow for handling submissions, which includes running an automated Airflow process that retrieves submissions and updates their status before piping them over to the Nextflow workflow.

### solution
- [x] Add a method to get submissions of a desired submission status (default is `RECEIVED`)
- [x] Add a method to update submission statuses of input submission(s)
- [x] Add tests
- [x] Add `challengeutils` as a dependency
- [x] Add documentation where appropriate
- [x] Fix failing CI jobs

### testing & preview 

Unit tests were added to `test_ops.py` to cover basic functionality and error-handling.

Further, here are some screenshots from actual communications with the API:

image
image

image
image


src/orca/services/synapse/ops.py
```diff
@@ -57,3 +58,73 @@ def monitor_evaluation_queue(self, evaluation_id: str) -> bool:
         )
         submissions_num = sum(1 for submission in received_submissions)
         return submissions_num > 0
+
+    def trigger_indexing(self, syn, submission_view) -> None:
```
thomasyu888: Nit: this is a general function for views (submission, file, entity, etc).  Type hinting.

src/orca/services/synapse/ops.py
```diff
@@ -1,8 +1,9 @@
 import logging
 from dataclasses import field
 from functools import cached_property
-from typing import ClassVar
+from typing import ClassVar, List, Union
 
+from challengeutils import utils
```
thomasyu888, line 9: This package is going to be deprecated in favor of cnb-tools eventually. For now it's ok, but the package isn't being regularly maintained.

Probably need to add this to the pipfile.

Maybe some of the challenge module should be added to the python client.

thomasyu888 (commented):
Awesome - I know you're still actively working on it, so just some preliminary comments.

jaymedina:
Hi @thomasyu888 thanks for the comments. Since this PR is a blocker for ORCA-302, I'd like to give a status update:

I see no way of mocking the `get_submissions` test since it requires retrieving submissions from an actual submissions view, so I'm constructing a set of integration tests where I build up and tear down a Project with an Evaluation queue for testing. This is what I'm working on right now and will aim to wrap up before the end of the day. @BWMac

thomasyu888:
> Hi @thomasyu888 thanks for the comments. Since this PR is a blocker for ORCA-302, I'd like to give a status update:
> 
> I see no way of mocking the `get_submissions` test since it requires retrieving submissions from an actual submissions view, so I'm constructing a set of integration tests where I build up and tear down a Project with an Evaluation queue for testing. This is what I'm working on right now and will aim to wrap up before the end of the day. @BWMac

See something like this: https://github.com/Sage-Bionetworks/Genie/blob/429dafbaed3bc78bbf71ff6bfffb7d88a6f49d4d/tests/test_clinical.py#L15-L46.  There is a way to create unit tests by mocking the behavior of `tableQuery` which may be lower lift than an integration test.  That said, those tests I wrote are a bit dated.  There are most likely other ways to achieve the mocking of a table query.

src/orca/services/synapse/ops.py
```diff
@@ -57,3 +61,77 @@ def monitor_evaluation_queue(self, evaluation_id: str) -> bool:
...
+        # Prepare a single submission_id for the for-loop
+        if isinstance(submission_ids, str):
+            submission_ids = [submission_ids]
+
+        # Let's catch for anything that was fed that is NOT a str or list of strs
+        if not isinstance(submission_ids, list):
+            raise TypeError(
+                "Not a list. ``submission_ids`` must be a string or list of strings."
+            )
+        if not all(isinstance(id, str) for id in submission_ids):
```
thomasyu888: This is interesting, submission ids could technically be integers, why strings?

.pre-commit-config.yaml
```diff
@@ -1,4 +1,4 @@
-exclude: '^docs/conf.py'
+exclude: '^docs/conf.py|^tests/services/synapse/test_ops.py'
```
jaymedina: The error-handling unit tests feed in values with wrong datatypes as a test, which pre-commit did not like. In order to keep this code coverage for testing without significantly compromising the use of pre-commit, I modified the pre-commit config to ignore `test_ops.py`.

Further, someone who claims to be the creator of `pre-commit` suggests opting tests out of the pre-commit runs entirely ([here](https://stackoverflow.com/questions/72103405/excluding-pytest-files-marks-from-pre-commit-hooks)). Maybe a topic for discussion at one of our team meetings?

jaymedina:
I'm going to mark this as RFR to move things along even though codecov is giving me 75%; The code that it says [isn't being covered](https://app.codecov.io/gh/Sage-Bionetworks-Workflows/py-orca/pull/37?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=Sage-Bionetworks-Workflows#diff-c3JjL29yY2Evc2VydmljZXMvc3luYXBzZS9vcHMucHk=) is covered [here](https://github.com/Sage-Bionetworks-Workflows/py-orca/blob/8bc7713251e101683151fd11f6d6471118883530/tests/services/synapse/test_ops.py#L163) and [here](https://github.com/Sage-Bionetworks-Workflows/py-orca/blob/8bc7713251e101683151fd11f6d6471118883530/tests/services/synapse/test_ops.py#L140) to my understanding.

I'll continue to look at this as I welcome review comments.

.pre-commit-config.yaml
BWMac: In that Stack Overflow response I think he was saying that you shouldn't _run_ your tests (i.e. execute the command `pytest tests`) as part of your `pre-commit` pipeline. Not that you shouldn't scan your testing scripts with other hooks.

src/orca/services/synapse/ops.py
```diff
@@ -57,3 +61,73 @@ def monitor_evaluation_queue(self, evaluation_id: str) -> bool:
...
+        """
+        Trigger indexing in Synapse.
+
+        Arguments:
+            syn: A Synapse object.
+            synapse_view: The Synapse ID of the view.
+        """
+        syn.tableQuery(f"select * from {synapse_view} limit 1")
+
+    def get_submissions(
```
BWMac: NIT: Should this have a slightly more specific name? `get_submissions_with_status`, for example.

src/orca/services/synapse/ops.py
```diff
@@ -57,3 +61,73 @@ def monitor_evaluation_queue(self, evaluation_id: str) -> bool:
...
+        self, submission_view: str, submission_status: str = "RECEIVED"
+    ) -> List[str]:
+        """
+        Get all submissions with desired submission status in a Synapse
+        submission view.
+
+        Arguments:
+            submission_view: The Synapse view to get submissions from.
```
BWMac: Specify that this takes the Synapse ID of the view.

src/orca/services/synapse/ops.py
```diff
@@ -57,3 +61,73 @@ def monitor_evaluation_queue(self, evaluation_id: str) -> bool:
...
+
+    def trigger_indexing(self, syn: Synapse, synapse_view: str) -> None:
+        """
+        Trigger indexing in Synapse.
+
+        Arguments:
+            syn: A Synapse object.
+            synapse_view: The Synapse ID of the view.
+        """
+        syn.tableQuery(f"select * from {synapse_view} limit 1")
```
BWMac: We don't need to take `syn` as an argument. Just use `self.client`

src/orca/services/synapse/ops.py
```diff
@@ -57,3 +61,73 @@ def monitor_evaluation_queue(self, evaluation_id: str) -> bool:
...
+            submission_view: The Synapse view to get submissions from.
+            submission_status: The submission status to filter for.
+                               Defaults to "RECEIVED".
+
+        Returns:
+            A list of submission IDs.
+        """
+
+        # Gain access to Synapse API
+        syn = self.client
```
BWMac: NIT: This is fine, but not necessary. I would just use `self.client`

.pre-commit-config.yaml
BWMac: Also, if we are going to exclude these would it be possible to exclude them from `mypy` only?

tests/services/synapse/test_ops.py
```diff
@@ -34,3 +68,155 @@ def test_monitor_evaluation_queue_returns_true_when_there_are_new_submissions(
...
+    raises an error if the input is neither a string nor a list.
+
+    Arguments:
+        mocker: A mocker object.
+        mocked_ops: A mocked instance of ``SynapseOps``.
+
+    """
+    # Test for non-string + non-list submission_ids.
+    error = "``submission_ids`` must be a string or list of strings."
+    with pytest.raises(TypeError) as err:
```
BWMac, line 197: I think this way of doing this is a little simpler

BWMac: I think this way of doing this is a little simpler: https://github.com/Sage-Bionetworks/agora-data-tools/blob/efe6eecf4acb0b5e1f4ff4c71ec0c59051a9c373/tests/test_utils.py#L38

tests/services/synapse/test_ops.py
```diff
@@ -34,3 +68,155 @@ def test_monitor_evaluation_queue_returns_true_when_there_are_new_submissions(
...
+    mock_update_status.assert_called_once_with("submission_1", "RECEIVED")
+
+
+def test_update_submission_status_with_non_string_non_list_input() -> None:
+    """
+    Tests that the ``update_submissions_status`` method in ``SynapseOps``
+    raises an error if the input is neither a string nor a list.
+
+    Arguments:
+        mocker: A mocker object.
```
BWMac: Are these arguments supposed to be passed to the test? You aren't mocking anything in this test and the one below

tests/services/synapse/test_ops.py
jaymedina: Ah, I forgot to comment on this one: So the reason I chose not to use `match` is so that I can handle the error message string-matching with my own assertion call and therefore write a more contextualized error message instead of the generic `couldn't match to Regex` assertion error.

What do you think of this? If we think the custom error message is not worth the extra lines I can use `match` instead.

tests/services/synapse/test_ops.py
```diff
@@ -34,3 +68,155 @@ def test_monitor_evaluation_queue_returns_true_when_there_are_new_submissions(
...
+    Tests that the ``update_submissions_status`` method in ``SynapseOps``
+    updates the status of one or more submissions in Synapse.
+
+    Arguments:
+        mocker: A mocker object.
+        mocked_ops: A mocked instance of ``SynapseOps``.
+
+    """
+    # Mocking the ``update_submission_status`` call in ``SynapseOps``
+    mock_update_status = mocker.patch.object(mocked_ops, "update_submissions_status")
```
BWMac, line 176: You shouldn't be mocking the function you are testing, you should be mocking dependent functions within it

tests/services/synapse/test_ops.py
```diff
@@ -34,3 +68,155 @@ def test_monitor_evaluation_queue_returns_true_when_there_are_new_submissions(
...
+
+    Arguments:
+        mocker: A mocker object.
+        mocked_ops: A mocked instance of ``SynapseOps``.
+    """
+    # Mocking the ``update_submission_status`` call in ``SynapseOps``
+    mock_update_status = mocker.patch.object(mocked_ops, "update_submissions_status")
+
+    # Calling the function to be tested with an input list of strings
+    mocked_ops.update_submissions_status(["submission_1", "submission_2"], "SCORED")
```
BWMac, line 155: You shouldn't be mocking the function you

tests/services/synapse/test_ops.py
```diff
@@ -34,3 +68,155 @@ def test_monitor_evaluation_queue_returns_true_when_there_are_new_submissions(
...
+    """
+    Tests that the ``update_submissions_status`` method in ``SynapseOps``
+    updates the status of one or more submissions in Synapse.
+
+    Arguments:
+        mocker: A mocker object.
+        mocked_ops: A mocked instance of ``SynapseOps``.
+    """
+    # Mocking the ``update_submission_status`` call in ``SynapseOps``
+    mock_update_status = mocker.patch.object(mocked_ops, "update_submissions_status")
```
BWMac: You shouldn't be mocking the function you are testing, just the dependent functions that are called within it

BWMac (changes requested):
My summary review comment got lost somehow, but I think your tests need some work. See my comments about mocking. I like to use the `patch` [method](https://docs.python.org/3/library/unittest.mock.html#patch) from `unittest.mock`. `agora-data-tools` has some [examples](https://github.com/Sage-Bionetworks/agora-data-tools/blob/efe6eecf4acb0b5e1f4ff4c71ec0c59051a9c373/tests/test_extract.py#L22) that you can reference.

src/orca/services/synapse/ops.py
```diff
@@ -57,3 +61,73 @@ def monitor_evaluation_queue(self, evaluation_id: str) -> bool:
...
+        # Get all submissions for the given ``submission_view``
+        query_results = syn.tableQuery(
+            f"select * from {submission_view} where status = '{submission_status}'"
+        )
+
+        submission_ids = query_results.asDataFrame()["id"].tolist()
+
+        return submission_ids
+
+    def update_submissions_status(
```
BWMac, line 106: For simplicity, should this only update one submission status? We can then run this in a loop in Airflow.

BWMac: For simplicity should this only take lists of IDs? We can always provide a list of one if need be. Or just take the individual ID and we can run the loop in Airflow

tests/services/synapse/test_ops.py
BWMac: I don't really think we need to worry about this. When an assertion fails in a test it tells you the expected and actual values in the stdout

src/orca/services/synapse/ops.py
jaymedina: Yes I think it would be good to limit the scope of this function and run the loop outside of it; have it run in airflow. If this works for you I'll go ahead and update this along with its tests.

BWMac: Sounds good!

tests/services/synapse/test_ops.py
BWMac: This is also why you aren't getting credit for the codecov. I think it looks for the tested function to be called but since you are mocking it, it doesn't count

jaymedina: Yep that's right - I have a bundle of commits coming up that I ran tests on locally and I'm hitting 90%. Just finishing up one last test.

tests/services/synapse/test_ops.py
jaymedina: No these were never args in here, this is a relic of copy/pasta

.pre-commit-config.yaml
jaymedina: Hmm this doesn't seem to be an issue anymore? I just ran precommit without this and it runs fine. Initially you'd think it's because I rescoped this func to only handle 1 submission at a time, and so the type hint changed to `Union[int, str]`, but I still have a unit test where I intentionally feed it a wrong data type (float, list, tuple). I've removed this exclusion from the config and will try to figure out why this happened.

src/orca/services/synapse/ops.py
```diff
@@ -57,3 +61,64 @@ def monitor_evaluation_queue(self, evaluation_id: str) -> bool:
...
+        """
+        Trigger indexing in Synapse.
+
+        Arguments:
+            syn: A Synapse object.
+            synapse_view: The Synapse ID of the view.
+        """
+        self.client.tableQuery(f"select * from {synapse_view} limit 1")
+
+    def get_submissions_with_status(
```
BWMac, line 75: Does this return an empty list when there are no submissions with the matching status? 

edit: NVM, I forgot that this was shown in the screenshots in the description

jaymedina: Actually it wasn't in the original screenshots - I've just updated them in the description box for record-keeping. Here is a screenshot answering your question. If none of the submissions have the status fed to `get_submissions_with_status` the list returned is empty and basically nothing happens. I'm going to add some error handling to catch typos/incorrect status types, but for status types [that do exist](https://rest-docs.synapse.org/rest/org/sagebionetworks/evaluation/model/SubmissionStatusEnum.html) but none of the submissions happen to be labeled with it, do we want some error-handling to catch this and break the Airflow run? Or do we want Airflow to quietly stop?

image

BWMac: My plan was for an empty list returned to lead to the `stop_dag` [task](https://github.com/Sage-Bionetworks-Workflows/orca-recipes/blob/4e6a92892ff5381c5d979b622f9f1de52ba64243/dags/challenge-submission-dag.py#L51) from a branching task. So it will quietly stop and we would expect this to happen very frequently if the DAG runs once/minute.

 As for error handling for incorrect `status` values, the Synapse API already does a decent job handling this with an enum:
```
challengeutils change-status 9742660 ABC

...

> synapseclient.core.exceptions.SynapseHTTPError: 400 Client Error:
> No enum constant org.sagebionetworks.evaluation.model.SubmissionStatusEnum.ABC
```

jaymedina: Sounds good, then I will leave the logic as it is for statuses that do exist in [here](https://rest-docs.synapse.org/rest/org/sagebionetworks/evaluation/model/SubmissionStatusEnum.html) but not assigned to any submission in the designated view.

I agree the error-handling for `update_submission_status` looks fine (wraps `challengeutils`). I've added error-handling for `get_submissions_with_status` since that was returning an empty list even if the status was just not a valid status, so I figured the extra error handling would be more helpful in tracking down this kind of problem if you ever have to. See my changes here https://github.com/Sage-Bionetworks-Workflows/py-orca/pull/37/commits/085837376f6cf7145440d37dc2a5f65488518a72

BWMac: Oh, I see what you are saying. Cool, I'll take a closer look when you are ready for the next round of review

jaymedina: Yep I'm ready

setup.cfg
```diff
@@ -68,6 +68,7 @@ exclude =
 
 # Additional dependencies for specific services
 all =
+    challengeutils~=4.3
     synapseclient~=2.7
```
BWMac, line 72: Can we bump this to the latest version?

src/orca/services/synapse/ops.py
```diff
@@ -57,3 +61,101 @@ def monitor_evaluation_queue(self, evaluation_id: str) -> bool:
...
+
+        Arguments:
+            submission_status: The submission status to check.
+
+        Returns:
+            Boolean indicating if submission status is valid.
+        """
+
+        # List all the available status options based on API docs
+        status_options = [
```
BWMac: I don't love that we are basically re-implementing the Synapse API enum here. I also am not sure this function is necessary at all considering `get_submissions_with_status` is a table query function. You can query the table with any string you want, and if that string is incorrect you should receive no responses since that is what the query would return. Do you feel strongly that this needs to be included?

setup.cfg
jaymedina: This is creating a dependency conflict that I'm fiddling with.

BWMac: OK, don't spend too much time on it if its being a pain. we can always create a separate ticket

src/orca/services/synapse/ops.py
BWMac: If you do think this should remain as part of this work, I think we should implement this as an [enum](https://docs.python.org/3/library/enum.html) of our own. 

In the future it should probably be added to the `synapseclient` as part of a `Submissions` OOP model.

setup.cfg
jaymedina: Yeah I'll make a separate ticket for this for the sake of moving this along.

src/orca/services/synapse/ops.py
jaymedina: I had the same frame of mind - my original plan was to pull the available statuses from the API but I found no way of doing that, so I went with recreating the list, but it does make this a little flimsy.

I don't feel strongly about keeping this function, but I do feel strongly about not letting a typo in the status argument get passed along quietly. At the very least, I think it would be good to add a warning in `get_submissions_with_status` when a list is empty, like `warnings.warn("This query returned an empty list. Ensure ``submission_status`` is correct")` to help catch potential mistakes. What do you think of this compromise?

BWMac: I like that even less since most of the time this function is called we expect this empty list to be returned so that warning will be all over the Airflow logs.

I think we either do it with the enum, or don't handle this at all.

jaymedina: Fair enough - In that case, let's not handle this just yet: I'll make a ticket to create an Enum for this in the python client, and one for revisiting this type of input error after the Enum gets created. I'm going to revert these changes in the next commit.

src/orca/services/synapse/ops.py
```diff
@@ -57,3 +61,67 @@ def monitor_evaluation_queue(self, evaluation_id: str) -> bool:
...
+
+        return submission_ids
+
+    def update_submission_status(
+        self,
+        submission_id: Union[int, str],
+        submission_status: str,
+    ) -> None:
+        """
+        Update the status of one or more submissions in Synapse.
```
BWMac, line 110:
```suggestion
        Update the status a submissions in Synapse.
```

BWMac:
```suggestion
        Update the status a submission in Synapse.
```

BWMac (approved):
LGTM. Thanks! Let me know when the new version has been released.

PR #40, opened 2024-04-09
[ORCA-308] Bumps `synapseclient` and `fs-synapse` versions & Removes `pipenv` files

BWMac:
**Description:**

This PR bumps up the `synapseclient` and `fs-synapse` versions that `py-orca` depends on. 

I also removed `Pipfile` and `Pipfile.lock` from the repository. This should have been done as part of #32. Because `pipenv` does not support constraints files it is impossible to guarantee a consistent installation of `apache-airflow` in the `py-orca` development environment. `py-orca` contributors should rely upon `dev_setup.sh` to [create a virtual environment](https://github.com/Sage-Bionetworks-Workflows/py-orca/blob/321de6c8fa711516af6c986a03d7141184ba1de8/CONTRIBUTING.md#L90-L91) on their local machines instead.


thomasyu888 (approved):
🔥 LGTM!

PR #42, opened 2024-04-19
[ORCA-318] Add instructions on setting up env on EC2

danlu1:
Problem:

The instruction doesn't include the update of the synindex workflow. It doesn't have guides on setting up the environment on EC2 and lacks details on AWS_Profile that can be used to run `demo.py`.

Solution:

Fill in more details in README.md.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ This Python package provides the components to connect various third-party servi
 
 ## Demonstration Script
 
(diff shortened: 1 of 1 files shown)


jaymedina (approved):
LGTM! 🚀 pre-approving but this looks good to merge once tests pass

BWMac (approved):
Thanks for adding this!

PR #44, opened 2024-06-17
Adds `gh_token` to KeepAlive workflow

BWMac:
Recently we had a [problem](https://github.com/Sage-Bionetworks/fs-synapse/pull/4) with `fs-synapse` where the `keepalive-workflow` did not have proper permissions to run when CI runs to prevent the workflow from being paused. This has not become an issue in this repo yet, but I want to prevent any future problems.


jaymedina (approved):
LGTM! Approved pending the final test passes

PR #45, opened 2024-10-28
[IBCDPE-1111] Upgrade airflow version

BryanFauble:
**Problem:**

1. In order to upgrade our apache airflow instance the version of airflow needs to be updated here.

**Solution:**

1. Upgrading the airflow dependency

**Testing:**

1. I tested this by creating a new apache airflow image and deploying it the test kubernetes cluster to verify DAGs were still able to run

Related PRs:

1. https://github.com/Sage-Bionetworks-Workflows/orca-recipes/pull/71
2. https://github.com/Sage-Bionetworks-Workflows/eks-stack/pull/42

### DIFF
--- a/requirements-airflow.txt
+++ b/requirements-airflow.txt
@@ -1,2 +1,2 @@
--c https://raw.githubusercontent.com/apache/airflow/constraints-2.7.2/constraints-3.10.txt
-apache-airflow==2.7.2
+-c https://raw.githubusercontent.com/apache/airflow/constraints-2.9.3/constraints-3.10.txt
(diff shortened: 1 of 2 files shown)


thomasyu888 (approved):
🔥 LGTM. i think my biggest concern was the Python version it requires, but I think we're on Python 3.10?

BryanFauble:
> 🔥 LGTM. i think my biggest concern was the Python version it requires, but I think we're on Python 3.10?

Yes, we are running with python >=3.10

BWMac (approved):
🔥

PR #48, opened 2025-03-25
[DPE-1275] Upgrade to 2.10.5

BryanFauble:
# **Problem:**

- There are some issues we are seeing with the apache airflow scheduler not running at expected times or consistenly

# **Solution:**

- Upgrade apache airflow to the latest version to take into account and and all bug fixes

# **Testing:**

- Once a new image is created we will be deploying the new images out to the dev k8s cluster


thomasyu888 (approved):
🔥 LGTM!

PR #49, opened 2025-03-31
[DPE-1275] Upgrade to the newest fs-synapse

BryanFauble:
# **Problem:**

- There are some issues we are seeing with the apache airflow scheduler not running at expected times or consistently. fs-synapse is a transitive dependency that also needs to be upgrades.

# **Solution:**

- Upgrade to the latest fs-synapse

# **Testing:**

- Once a new image is created we will be deploying the new images out to the dev k8s cluster


thomasyu888 (approved):
🔥 LGTM!

PR #51, opened 2025-05-08
[DPE-1288] Updates synapseclient version

BWMac:
# **Problem:**

We need access to new OOP models such as `DatasetCollection` in `orca-recipes` DAGs.

# **Solution:**

Upgrade the `synapseclient` version in this repository.

# **Testing:**

Ci pipeline.

# **Notes:**

The old `gautamkrishnar/keepalive-workflow@2.0.10` workflow we used to use to prevent GH from shutting off our automated actions appears to have been [banned](https://github.com/gautamkrishnar/keepalive-workflow) from the platform due to some policy violation. I have replaced it with a different one that serves the same purpose, but will run each time our scheduled testing runs rather than every X number of days.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,6 +26,8 @@ concurrency:
 jobs:
   prepare:
     runs-on: ubuntu-latest
+    permissions:
+      actions: write
     outputs:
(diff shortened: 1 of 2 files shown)


thomasyu888 (approved):
🔥 LGTM!

PR #52, opened 2025-10-15
[DPE-1474] remove entry_name, add new profiles

jaymedina:
# **Problem:**

The current `demo.py` is outdated, following changes to the `entry` parameter, causing workflow runs to fail:

https://tower-dev.sagebionetworks.org/orgs/Sage-Bionetworks/workspaces/example-dev-project/watch/5FkvJwV3YCQMnf

# **Solution:**

Similar to [other references to Tower](https://github.com/Sage-Bionetworks-Workflows/orca-recipes/blob/69d1fe6bcda0eeeec0b7d6ac77991b5e5c96a855/dags/challenge_dag_factory.py#L285), remove `entry_param` from the LaunchInfo calls, and add the [profiles](https://github.com/Sage-Bionetworks-Workflows/nf-synapse/blob/77293186c4ee3e74de98cc698bb17db125bc332e/nextflow.config#L15) that contain the new `params.entry` parameter introduced last year that replaces the `Workflow entry name` field in SP.

# **Testing:**

### DIFF
--- a/demo.py
+++ b/demo.py
@@ -46,8 +46,7 @@ class RnaseqDataset:
             run_name=run_name,
             pipeline="Sage-Bionetworks-Workflows/nf-synapse",
             revision="main",
-            profiles=["docker"],
-            entry_name="NF_SYNSTAGE",
+            profiles=["docker", "synstage"],
             params={
                 "input": samplesheet_uri,
             },
@@ -97,8 +96,7 @@ class RnaseqDataset:
             run_name=self.get_run_name("synindex"),
             pipeline="Sage-Bionetworks-Workflows/nf-synapse",
             revision="main",
-            profiles=["docker"],
-            entry_name="NF_SYNINDEX",
+            profiles=["docker", "synindex"],
             params={
                 "s3_prefix": rnaseq_outdir_uri,
                 "parent_id": self.output_folder,


rxu17 (approved):
LGTM!

demo.py
rxu17, line 1: I wonder if we should just be removing `entry_name` option from `LaunchInfo` here since it's deprecated in Nextflow and we're going to use `profiles` or custom `entry` param/depends on the nf workflow: https://github.com/Sage-Bionetworks-Workflows/py-orca/blob/7e54c1572c5287c036fbb31774f304a11e9080e3/src/orca/services/nextflowtower/models.py#L170

jaymedina: Good catch, this should also be removed

Edit: I scouted around `orca-recipes` to see if anything else still uses `entry_name` that this change would break, and I saw some stuff come up:

https://github.com/search?q=repo%3ASage-Bionetworks-Workflows%2Forca-recipes+entry_name&type=code

So I wonder if we should make a formal announcement on one of the Tower channels asking people to update their pipeline launches before we make this sort of change

rxu17: Yes, I agree with the formal announcement since we'll be deprecating a major tower launch parameter, and linking the list you found of potentially still active workflows that are using (and there may be more).

jaymedina: Sounds good, I agree! I've made a ticket to announce this to users + officially remove the `entry_name` parameter: https://sagebionetworks.jira.com/browse/DPE-1484

